PR #37, opened 2022-09-14
fix: valid chunk count

metacertain:


src/PostageStamp.sol
```diff
@@ -173,6 +173,8 @@ contract PostageStamp is AccessControl, Pausable {
 
         uint256 normalisedBalance = currentTotalOutPayment() + (_initialBalancePerChunk);
```
istae, line 174: the normalisedBalance require line on 184 should be moved to 175. The transfer of tokens should be made after the last require imo. The samething should be applied to createBatch function as well.

istae (approved):

PR #42, opened 2022-10-24
feat: draft

metacertain:


ldeffenb:
Was this perchance triggered by the claim in round 51683?  https://github.com/rndlabs/monSI/issues/26

src/Redistribution.sol
```diff
@@ -76,9 +79,10 @@ contract Redistribution is AccessControl, Pausable {
     /**
      * @param staking the registry used by this contract
      */
-    constructor(address staking, address postageContract) {
+    constructor(address staking, address postageContract, address oracleContract) {
```
ldeffenb, line 82: Should the comment be updated to reflect the added parameters?

metacertain:
> Was this perchance triggered by the claim in round 51683? [rndlabs/monSI#26](https://github.com/rndlabs/monSI/issues/26)

Your issue sounds like the claim of the previous round was delayed so much on goerli that it was only processed in the next claim phase, and became a valid transaction (but still the payout went to the next round winner, so it didn't change the outcome)

ldeffenb:
Yes, that's the conclusion I came to as well.  And while the distribution was actually correct, it seems unfair that the wrong overlay paid the fee.   I see there's a round number added to the commit() which is good as it prevents delayed commits from possibly causing a slash, but wonder if the same addition and check should be added to the claim()?

PR #52, opened 2022-11-09
feat: single expire function

istae:
The trick here is that by updating lastExpiryBalance when there are no more expired batches left in the loop,
it has the same behavior of the old `expire` function

added benefit is the batches can be expired in increments while the pot is incremented properly.

### DIFF
--- a/src/PostageStamp.sol
+++ b/src/PostageStamp.sol
@@ -361,7 +361,7 @@ contract PostageStamp is AccessControl, Pausable {
             if (empty()) {
                 lastExpiryBalance = currentTotalOutPayment();
                 break;
-            } 
+            }
             // get the batch with the smallest normalised balance
             bytes32 fbi = firstBatchId();
             // if the batch with the smallest balance has not yet expired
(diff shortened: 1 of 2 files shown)


src/PostageStamp.sol
```diff
@@ -382,43 +392,22 @@ contract PostageStamp is AccessControl, Pausable {
...
      */
-    function hasExpired() public view returns(bool) {
-        bytes32 fbi = firstBatchId();
-        if (remainingBalance(fbi) > 0) {
+    function expiredBatchesExist() public view returns (bool) {
+        if (empty()){
             return false;
         }
-        return true;
+        return !(remainingBalance(firstBatchId()) > 0);
```
istae, line 403: this is funky :)

just do `return remainingBalance(firstBatchId()) < 0`

src/Redistribution.sol
```diff
@@ -9,93 +9,70 @@ import "./Staking.sol";
...
- * During the _reveal_ round, the randomness is updated after every successful reveal. Once the reveal round
- * is concluded, the _currentRoundAnchor_ is updated and users can determine if they will be eligible their overlay
- * will be eligible for the next commit phase using _isParticipatingInUpcomingRound_.
- *
- * When the _reveal_ phase has been concluded, the claim phase can begin. At this point the truth teller and winner
- * are already determined. By calling _isWinner_, an applicant node can run the relevant logic do determine if they have
- * been selected as the beneficiary of this round. When calling _claim_, the current pot from the PostageStamp contract
- * is withdrawn and transferred to the beneficiaries address. Nodes that have revealed values that differ from the truth,
- * have their stakes "frozen" for a period of rounds proportional to their reported depth.
+ * @dev The redistribution contracts allows users to create and manage postage stamp batches.
```
istae, line 12: your PR pulled in some random commit, remove the changes to src/Redistribution.sol

istae: Please remember to self review your changes/PRs

src/PostageStamp.sol
ldeffenb: ! > 0 is not the same as < 0, but actually <= 0, if zero matters.

PR #55, opened 2022-11-17
feat: add staking and redistribution go generation

mrekucci:
Add staking and redistribution go generation plus some cleanups.


istae (approved):
beautiful

PR #59, opened 2022-11-23
feat: added script for one click deployment

swarmHaseeb:
description: script for deployment
further improvements (in next PR): documentation + script to use the deployed.json files to run a bee node with appropriate changes.


deploy/deploy_staking.ts
```diff
@@ -0,0 +1,92 @@
+import { ethers } from 'hardhat';
+import { HardhatRuntimeEnvironment } from 'hardhat/types';
+import { DeployFunction } from 'hardhat-deploy/types';
+import 'hardhat-deploy-ethers';
+import hre from 'hardhat';
+
+import * as addresses from '../deployed.json';
+import { ContractReceipt, ContractTransaction } from 'ethers';
+
+const deployTo = process.env.DEPLOYTO;
```
mrekucci, line 10: I'd rename the env and the const to `blockChainVendor`.

deployed.json
```diff
@@ -0,0 +1,40 @@
+{
```
mrekucci, line 1: I'd recommend the following fields:
```json
{
  "abi": "",
  "bytecode": "",
  "deployedBytecode": "",
  "address": "",
  "url": "",
}
```
Not sure what the "deployer" field should hold?

hardhat.config.ts
```diff
@@ -14,6 +14,14 @@ if (!mnemonic) {
   mnemonic = 'test test test test test test test test test test test test';
 }
 
+let MNEMONICORKEY = process.env.MNEMONICORKEY;
```
mrekucci, line 17: Why do we need both `mnemonic ` and `MNEMONICORKEY`?

deploy/deploy_staking.ts
```diff
@@ -0,0 +1,92 @@
...
+const networkID = 5; //test network
```
mrekucci, line 11: The network ID should be given as a parameter to this script since we'll deploy to several networks.

deploy/deploy_staking.ts
```diff
@@ -0,0 +1,92 @@
...
+
+// all contract addresses will be set here
+let factory = '';
```
mrekucci, line 14: These values should be read from the `deployed.json` file.

swarmHaseeb:
@agazso for kind notice.

scripts/deployer.ts
```diff
@@ -0,0 +1,197 @@
...
+import 'hardhat-deploy-ethers';
+import hre from 'hardhat';
+import * as fs from 'fs';
+import { ContractReceipt, ContractTransaction } from 'ethers';
+import mainnetData from '../mainnet_deployed.json';
+import testnetData from '../testnet_deployed.json';
+import dummyData from '../example_testnet_deployed.json';
+import { getCurrentTimestamp } from 'hardhat/internal/hardhat-network/provider/utils/getCurrentTimestamp';
+
+let deployedData = dummyData;
```
AuHau, line 13: The `mainnetData`, `testnetDat` and `dummyData` share the same structure right? I would propose creating Interface with this structure which you would declare here and then you should not need the `@ts-ignore` bellow when you reassign it with mainnet/testnet data...

Please double check what I wrote, most probably there is some mistake somewhere as I am not writing it in IDE 😅

```suggestion

interface DeployedContract {
      abi: Array<any>, // Improve these types
      bytecode: string,
      address: string,
      block: number,
      url: string
} 
interface DeployedData {
	contracts: {
		postage: DeployedContract,
		redistribution: DeployedContract,
		staking: DeployedContract,
		priceOracle: DeployedContract,
		factory: DeployedContract
	}
}

let deployedData: DeployedData = dummyData;
```

scripts/deployer.ts
```diff
@@ -0,0 +1,197 @@
...
+let postage = '';
+let oracle = '';
+let redistribution = '';
+
+// all block numbers will be set here
+let factoryBlock = 0;
+let stakingBlock = 0;
+let postageBlock = 0;
+let oracleBlock = 0;
+let redistributionBlock = 0;
```
AuHau, line 40: This is not really good approach 😅 I would instead suggest passing these values as function parameters instead of having them as "global variables". Simply construct an object that will contain these properties as you deploy them and then pass it into the functions that process them.

scripts/deployer.ts
```diff
@@ -0,0 +1,197 @@
...
+
+async function main() {
+  if (blockChainVendor != 'testnet' && blockChainVendor != 'mainnet') {
+    const TestToken = await hre.ethers.getContractFactory('TestToken');
+    const testToken = await TestToken.deploy();
+    await testToken.deployed();
+    factory = testToken.address;
+    factoryBlock = testToken.deployTransaction.blockNumber as number;
+  } else if (blockChainVendor == 'testnet') {
+    const deployed = await JSON.parse(JSON.stringify(testnetData).toString());
```
AuHau, line 50: Why do you use this construct? IMHO you should not need it...

Generally, this is used to make a deep copy of an object, but it is rarely used and usually signals some bad approach.

scripts/deployer.ts
AuHau: If the problem is that you do not have all the info at the same time then you can use [`Partial<>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype) helper type that allows all the direct properties of the type to be `undefined`.

mrekucci (commented):
The `example_testnet_deployed.json` is non needed.

scripts/deployer.ts
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  const StakeRegistry = await hre.ethers.getContractFactory('StakeRegistry');
+  const stakeRegistry = await StakeRegistry.deploy(factory, networkID);
+  await stakeRegistry.deployed();
+  stakingBlock = stakeRegistry.deployTransaction.blockNumber as number;
+  staking = stakeRegistry.address;
+
+  const Postage = await hre.ethers.getContractFactory('PostageStamp');
+  const postageadd = await Postage.deploy(factory);
```
mrekucci, line 76: `postageadd` -> `postageContract`

scripts/deployer.ts
```diff
@@ -0,0 +1,197 @@
...
+  await postageadd.deployed();
+  postageBlock = postageadd.deployTransaction.blockNumber as number;
+  postage = postageadd.address;
+
+  const Oracle = await hre.ethers.getContractFactory('PriceOracle');
+  const oracleAdd = await Oracle.deploy(postage);
```
mrekucci, line 82: `oracleAdd` -> `oracleContract`

scripts/deployer.ts
```diff
@@ -0,0 +1,197 @@
...
+  await oracleAdd.deployed();
+  oracleBlock = oracleAdd.deployTransaction.blockNumber as number;
+  oracle = oracleAdd.address;
+
+  const Redistribution = await hre.ethers.getContractFactory('Redistribution');
+  const redisAdd = await Redistribution.deploy(staking, postage, oracle);
```
mrekucci, line 88: `redisAdd` -> `redistributionContract`

scripts/deployer.ts
```diff
@@ -0,0 +1,197 @@
...
+    // @ts-ignore
+    deployedData = mainnetData;
+  }
+
+  if (!factory) {
+    handleErr('expected factory address');
+  }
+
+  const StakeRegistry = await hre.ethers.getContractFactory('StakeRegistry');
+  const stakeRegistry = await StakeRegistry.deploy(factory, networkID);
```
mrekucci, line 70: `stakeRegistry` -> `stakeRegistryContract`

scripts/deployer.ts
swarmHaseeb: it also looked funny btw :laughing:

hardhat.config.ts
```diff
@@ -112,11 +111,23 @@ const config: HardhatUserConfig = {
...
+      url: 'http://127.0.0.1:8545',
+      chainId: 1337,
+    },
+    testnet: {
+      url: 'https://goerli.infura.io/v3/' + process.env.INFURA_TOKEN,
+      accounts: [process.env.WALLET_SECRET as string],
+      chainId: 5,
+      gasPrice: 80000000000, // this is 80 Gwei
+    },
   },
```
mrekucci, line 131: The `mainnet` network profile should also be declared here.

scripts/deployer.ts
```diff
@@ -0,0 +1,279 @@
...
+import { ContractReceipt, ContractTransaction } from 'ethers';
+import mainnetData from '../mainnet_deployed.json';
+import testnetData from '../testnet_deployed.json';
+import dummyData from '../example_testnet_deployed.json';
+import { getCurrentTimestamp } from 'hardhat/internal/hardhat-network/provider/utils/getCurrentTimestamp';
+import { ethers } from 'hardhat';
+import hre from 'hardhat';
+
+interface DeployedContract {
+  abi: Array<any>; // Improve these types
```
mrekucci, line 12: This lint warning should be fixed.

scripts/deployer.ts
```diff
@@ -0,0 +1,279 @@
...
+  const StakeReg = await ethers.getContractFactory('StakeRegistry');
+
+  const contract = PostageStamp.attach(contractData.addresses.postageStamp);
+
+  let result: ContractTransaction;
+  let receipt: ContractReceipt;
+
+  const redistributorRole = contract.REDISTRIBUTOR_ROLE();
+  result = await contract.grantRole(redistributorRole, contractData.addresses.redistribution);
+  receipt = await result.wait();
```
mrekucci, line 193: The `receipt` is never used: `receipt = await result.wait();` -> `await result.wait();`

scripts/deployer.ts
```diff
@@ -0,0 +1,279 @@
...
+
+  await rolesSetter(contractData);
+
+  await writeResult(deployedData, contractData);
+  await logResult(contractData);
+}
+
+async function deployFactory(deployed: any, contractData: any) {
+  if (deployed == null) {
+    const TestToken = await hre.ethers.getContractFactory('TestToken');
```
mrekucci, line 89: Not sure about this, is this correct when deploying to the mainnet?

mrekucci (commented):
The `example_testnet_deployed.json` shouldn't be included.

.gitignore
```diff
@@ -22,6 +22,12 @@ node_modules/
 deployments/hardhat
 deployments/localhost
 
+# deployment files for mainnet and testnet and an example file
+*_deployed.json
+!mainnet_deployed.json
+!testnet_deployed.json
+!example_testnet_deployed.json
```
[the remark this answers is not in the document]
swarmHaseeb: it should be included because it is default file for other nodes and the testnet/mainnet files are not added in git ignore. It is a safety pin.

scripts/deployer.ts
swarmHaseeb: removed the comment, this any tag is alright

scripts/deployer.ts
swarmHaseeb: yes, looks good

hardhat.config.ts
```diff
@@ -112,10 +112,26 @@ const config: HardhatUserConfig = {
...
+    },
+    testnet: {
       url: 'https://goerli.infura.io/v3/' + process.env.INFURA_TOKEN,
       accounts,
+      chainId: 5,
+    },
+    mainnet: {
+      url: 'https://mainnet.infura.io/v3/' + process.env.INFURA_TOKEN,
+      accounts,
+      chainId: 1,
```
mrekucci, line 134: Should be `100`.

mrekucci (approved):

mainnet_deployed.json
```diff
@@ -0,0 +1,39 @@
+{
```
istae, line 1: is this file supposed to be here?

swarmHaseeb: yes

testnet_deployed.json
```diff
@@ -0,0 +1,39 @@
+{
```
istae, line 1: same with this file? 
are these temporary files?

swarmHaseeb: nope, not temporary, they will be generated and filled from the node

hardhat.config.ts
```diff
@@ -112,10 +112,26 @@ const config: HardhatUserConfig = {
     localhost: {
       url: 'http://localhost:8545',
```
notanatol, line 113: let's change this to be `http://geth-swap.localhost` by default, since we will be using it against the `beelocal` cluster

notanatol (approved):

istae (approved):

hardhat.config.ts
```diff
@@ -5,12 +5,12 @@ import 'hardhat-deploy';
 import 'hardhat-deploy-ethers';
 import 'hardhat-tracer';
 
-// Define mnemonic for accounts.
 let mnemonic = process.env.MNEMONIC;
 if (!mnemonic) {
   // NOTE: this fallback is for development only!
   // When using other networks, set the secret in .env.
   // DO NOT commit or share your mnemonic with others!
+  console.log(mnemonic);
```
significance, line 13: is it needed to leak this?

scripts/deployer.ts
```diff
@@ -0,0 +1,280 @@
...
+    redistribution: DeployedContract;
+    staking: DeployedContract;
+    priceOracle: DeployedContract;
+    factory: DeployedContract;
+  };
+}
+
+// URL
+const mainnetURL = 'https://gnosisscan.io/address/';
+const testnetURL = 'https://goerli.etherscan.io/address/';
```
significance, line 30: can we add these into .env ?

significance: or fetch from hre config?

scripts/deployer.ts
significance: @mrekucci is correct i believe, we should not deploy the testToken when deploying to mainnet

significance: instead we should fetch the BZZ token address from .env

scripts/deployer.ts
```diff
@@ -0,0 +1,280 @@
...
+  contractData.addresses.factory = deployed['contracts']['factory']['address'];
+  contractData.blocks.factory = deployed['contracts']['factory']['block'];
+  return contractData;
+}
+
+async function deployStaking(contractData: any) {
+  const StakeRegistryContract = await hre.ethers.getContractFactory('StakeRegistry');
+  const stakeRegistryContract = await StakeRegistryContract.deploy(contractData.addresses.factory, networkID);
+  await stakeRegistryContract.deployed();
+  console.log(stakeRegistryContract.address);
```
significance, line 102: is it needed to log this? if so it would be good to be more descriptive

scripts/deployer.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  const contract = PostageStamp.attach(contractData.addresses.postageStamp);
+
+  let result: ContractTransaction;
+  let receipt: ContractReceipt;
+
+  const redistributorRole = contract.REDISTRIBUTOR_ROLE();
+  result = await contract.grantRole(redistributorRole, contractData.addresses.redistribution);
+  receipt = await result.wait();
+  console.log(receipt);
```
significance, line 189: same, please remove or make more descriptive

scripts/deployer.ts
```diff
@@ -0,0 +1,280 @@
...
+    deployed['contracts']['priceOracle']['url'] = urlAddress;
+  }
+
+  if (fileName.length == 0 || !fs.existsSync(fileName)) {
+    fileName = blockChainVendor + '_deployed.json';
+  }
+  fs.writeFileSync(fileName, JSON.stringify(deployed, null, '\t'));
+}
+
+// We recommend this pattern to be able to use async/await everywhere
```
significance, line 268: remove unneeded comments

significance (changes requested):
great start! thanks @swarmHaseeb 

as discussed, we will need to get into the detail of what needs to be done to relink/bootstrap the contracts, but let's finalise and merge this and we can augment this as we deal with those tasks

scripts/deployer.ts
mrekucci: I'd recommend putting this into the config.

scripts/deployer.ts
mrekucci: It'll be fetched from the `*_deployed.json` files once bootstrapped.

scripts/deployer.ts
mrekucci: Agree, this is not needed since it's logged below. The same applies to other functions.

scripts/deployer.ts
```diff
@@ -0,0 +1,280 @@
...
+      priceOracle: 0,
+      factory: 0,
+    },
+  };
+
+  switch (blockChainVendor) {
+    case 'testnet':
+      contractData = await deployFactory(await JSON.parse(JSON.stringify(testnetData).toString()), contractData);
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
AuHau, line 57: So IMHO these `ts-ignore`s are avoidable. I would suggest creating interface for `contractData` variable and then use it in its declaration. The form of the testnetData/mainnetData are then same right?

scripts/deployer.ts
```diff
@@ -0,0 +1,280 @@
...
+// We recommend this pattern to be able to use async/await everywhere
+// and properly handle errors.
+main()
+  .then(() => process.exit(0))
+  .catch((error) => {
+    console.error(error);
+    process.exit(1);
+  });
+
+function handleErr(err: string) {
```
AuHau, line 277: Honestly, I would suggest simply throwing an `Error` instead of this way of handling it. 

It has benefits like you see stack trace and IMHO is cleaner from readability as well, as you know what happens. Moreover, this exact handling is then performed on the top-level error handler [here](https://github.com/ethersphere/storage-incentives/pull/59/files#diff-21e04f146ffa17a95150f99a1ff0da362398c106a1a4b029a9e8dc4666113ec6R272)

AuHau (commented):
Good step forward, but IMHO still needs some tweaks ;-) 

Generally, avoid using `any` ;-) Instead if possible define the types or use `unknown` which forces you to do some checks when needed...

hardhat.config.ts
```diff
@@ -5,12 +5,12 @@ import 'hardhat-deploy';
 import 'hardhat-deploy-ethers';
 import 'hardhat-tracer';
 
-// Define mnemonic for accounts.
 let mnemonic = process.env.MNEMONIC;
```
mrekucci, line 8: Replace with:
```ts
const mnemonic = process.env.MNEMONIC;
if (!mnemonic) {
  throw new Error('Please set your MNEMONIC in a .env file');
}
const accounts = { mnemonic };

const infuraToken = process.env.INFURA_TOKEN;
if (!infuraToken) {
  throw new Error('Please set your INFURA_TOKEN in a .env file');
}
```

significance (approved):

PR #73, opened 2022-12-05
feat: set price oracle current price to its minimum price as default

### LINKED ISSUE #66: price oracle should have an initial starting price

mrekucci:
Closes #66


src/PriceOracle.sol
```diff
@@ -17,12 +17,12 @@ contract PriceOracle is AccessControl {
...
     bytes32 public constant PRICE_UPDATER_ROLE = keccak256("PRICE_UPDATER");
 
-    // The current price is the atomic unit
-    uint256 public currentPrice;
-
     // The minimum price allowed
     uint256 public constant minimumPrice = 1024;
 
+    // The current price is the atomic unit.
+    uint256 public currentPrice = minimumPrice;
```
swarmHaseeb, line 24: Maybe add a require condition within contract that ensures that the `currentPrice` is never less than `minimumPrice`

test/PriceOracle.test.ts
```diff
@@ -49,6 +49,13 @@ describe('PriceOracle', function () {
       const postageStamp = await priceOracle.postageStamp();
       expect(postageStamp).to.be.eq((await ethers.getContract('PostageStamp')).address);
     });
+
+    it('current price should be equal to minimum price', async function () {
```
swarmHaseeb, line 53: Maybe add a test case that ensures that the `currentPrice` is never less than `minimumPrice`.

swarmHaseeb (changes requested):

src/PriceOracle.sol
mrekucci: Can you be more specific? I'm not sure where you're proposing to add it. In both functions (`adjustPrice`, `setPrice`) this is already addressed.

test/PriceOracle.test.ts
mrekucci: Done.

src/PriceOracle.sol
swarmHaseeb: @significance @istae can you please maybe look into it and if its a bug we can create a new issue

test/PriceOracle.test.ts
swarmHaseeb: @significance @istae can you please maybe look into it and if its a bug we can create a new issue

istae (approved):

swarmHaseeb (approved):

PR #93, opened 2022-12-13
feat: add minimum batch depth

### LINKED ISSUE #91: Minimum bucket depth in the PostageStamp contract
Previously discussed here: #56

It turns out that the point to have a minimum bucket depth is to avoid having postage stamps with very small depth, therefore avoiding the problem of having a large number of low value postage stamps causing potential issues when managing them in the ordered tree in the contract.

The recommendation is to set the minimum bucket depth to 16 and add back checks so that the minimum postage batch depth is bigger than that.

Another issue was that there are integration tests in Bee that wants to use lower values so that testing can be done quickly. For that the idea is that the minimum bucket depth could be set at constructor time or set with a setter function with admin rights on the contract. Then the tests could set any arbitrary low value that would make them pass.

mrekucci:
Closes #91

### DIFF
--- a/src/PostageStamp.sol
+++ b/src/PostageStamp.sol
@@ -71,8 +71,10 @@ contract PostageStamp is AccessControl, Pausable {
     bytes32 public constant PRICE_ORACLE_ROLE = keccak256("PRICE_ORACLE");
     // Role allowed to pause
     bytes32 public constant PAUSER_ROLE = keccak256("PAUSER_ROLE");
-    // Role allowed to withdraw the pot
+    // Role allowed to withdraw the pot.
     bytes32 public constant REDISTRIBUTOR_ROLE = keccak256("REDISTRIBUTOR_ROLE");
+    // Role allowed to specify the minimum depth.
+    bytes32 public constant DEPTH_ORACLE_ROLE = keccak256("DEPTH_ORACLE_ROLE");
 
     // Associate every batch id with batch data.
     mapping(bytes32 => Batch) public batches;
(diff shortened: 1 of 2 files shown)


istae (approved):

src/PostageStamp.sol
```diff
@@ -430,4 +436,12 @@ contract PostageStamp is AccessControl, Pausable {
         require(ERC20(bzzToken).transferFrom(msg.sender, address(this), amount), "failed transfer");
         pot += amount;
     }
+
+    /**
+     * @notice Set minimum batch depth
+     */
+    function setMinimumBatchDepth(uint8 min) external {
```
istae, line 443: what about having this an a constructor arg?

mrekucci: Done.

agazso (changes requested):
Please note that the issue is about minimum **bucket** depth and not just about minimum batch depth. There is also a transitive requirement, such that the batch depth has to be greater than the bucket depth, therefore it has also be greater than the minimum bucket depth.

mrekucci:
> Please note that the issue is about minimum **bucket** depth and not just about minimum batch depth. There is also a transitive requirement, such that the batch depth has to be greater than the bucket depth, therefore it has also be greater than the minimum bucket depth.

Good catch! I restored what was removed [here](https://github.com/ethersphere/storage-incentives/commit/1619543fff1cdd14459210610086810991802aa9#diff-ca3aad745621d1cc8c0287779901e53d29d3d295caa604fee27369a1b8514661L91). Fixed.

agazso (approved):

istae (approved):

PR #94, opened 2022-12-13
restore constraint on diluting immutable batches

significance:


mrekucci:
Closed in favor of #93 and #96.

PR #104, opened 2022-12-18
feat: randomness method update

metacertain:


nagydani (approved):
Accepted.

nugaon (approved):
Seed can be updated only on the first reveal,
rid of mining potential on randomNumber update in the `isWinner` function.

PR #110, opened 2023-01-19
feat: Inclusion proofs and signature validation

metacertain:


mrekucci (commented):
Would be nice to have some tests.

src/PostageStamp.sol
```diff
@@ -93,6 +97,9 @@ contract PostageStamp is AccessControl, Pausable {
 
     // Lottery pot at last update.
     uint256 public pot;
+    
+    // blocks in 24 hours ~ 24 * 60 * 60 / 5 = 17280
```
significance, line 101: perhaps should add blocktime as a init var and calculate accordingly

src/PriceOracle.sol
```diff
@@ -24,7 +24,7 @@ contract PriceOracle is AccessControl {
     uint256 public currentPrice = minimumPrice;
 
     // Constants used to modulate the price, see below usage
-    uint256[] public increaseRate = [0, 1069, 1048, 1032, 1024, 1021, 1015, 1003, 980];
```
significance, line 27: we should address this in a separate PR

nugaon:
Continuing in #152

PR #111, opened 2023-01-31
feat: stamp updates

metacertain:


src/PriceOracle.sol
```diff
@@ -24,7 +24,7 @@ contract PriceOracle is AccessControl {
     uint256 public currentPrice = minimumPrice;
 
     // Constants used to modulate the price, see below usage
-    uint256[] public increaseRate = [0, 1069, 1048, 1032, 1024, 1021, 1015, 1003, 980];
+    uint256[] public increaseRate = [0, 1036, 1027, 1025, 1024, 1023, 1021, 1017, 1012];
```
nugaon, line 27: cannot review

nugaon (approved):

src/PostageStamp.sol
```diff
@@ -221,11 +232,11 @@ contract PostageStamp is AccessControl, Pausable {
         require(batch.owner != address(0), "batch does not exist or has expired");
         require(batch.normalisedBalance > currentTotalOutPayment(), "batch already expired");
         require(batch.depth > minimumBucketDepth, "batch too small to renew");
+        require(batch.normalisedBalance + (_topupAmountPerChunk) >= minimumInitialBalancePerChunk(), "insufficient topped up balance for 24h minimum validity");
```
nagydani, line 235: This is incorrect, because it allows for keeping a batch near expiry cheaply.

test/PostageStamp.test.ts
```diff
@@ -98,6 +100,8 @@ describe('PostageStamp', function () {
 
         setPrice0Block = await getBlockNumber();
         await priceOracle.setPrice(price0);
+        await priceOracle.setPrice(price0);
```
nagydani, line 103: Do not do this twice. Looks like an editing mistake.

nagydani (commented):
Some problems have been uncovered during the review, see inline comments. Please fix before approval.

src/PostageStamp.sol
```diff
@@ -61,10 +61,14 @@ contract PostageStamp is AccessControl, Pausable {
         address owner;
         // Current depth of this batch.
         uint8 depth;
+        //
+        uint8 bucketDepth;
```
zelig, line 65: i starting to doubt this was a good name suggests the log size of a bucket. eitger call it 
- `maxNeighbourhoodDepth` OR
- `minCollisionDepth`

nagydani (approved):
Changes approved. The rent oracle should be tamed a bit more...

src/PostageStamp.sol
```diff
@@ -93,6 +102,9 @@ contract PostageStamp is AccessControl, Pausable {
 
     // Lottery pot at last update.
     uint256 public pot;
+    
+    // blocks in 24 hours ~ 24 * 60 * 60 / 5 = 17280 
+    uint256 public minimumValidityBlocks = 17280;
```
ldeffenb, line 108: Isn't this a bit hard-coded for the gnosis/xdai chain?   This will be exceedingly (2.4*) long on the goerli test chain.

mfw78: This could be solved by:
1. Making the variable immutable, and then specifically passed through the constructor at contract deployment; or
2. Given that the nature of this is to determine a length in blocks, and we know that there are deployments to goerli, and gnosis chain, one could:

  a. Use assembly to retrieve the chain id.
  b. Based on the chain id, do the math and allocate to the immutable public variable during the constructor. eg:
  
  ```solidity
      constructor(address _bzzToken, uint8 _minimumBucketDepth) {
        bzzToken = _bzzToken;
        minimumBucketDepth = _minimumBucketDepth;
        _setupRole(DEFAULT_ADMIN_ROLE, msg.sender);
        _setupRole(PAUSER_ROLE, msg.sender);
        
        // Set chain-specific var
        uint256 chainId;
        assembly {
          chainId := chainid()
        }
        
        if (chainId == 5) minimumValidityBlocks = 24 hours / 12;
        if (chainId == 100) minimumValidityBlocks = 24 hours / 5;
    }
  ```
  
Given the low cost of deployments on Gnosis Chain, and zero cost on Goerli, this does present a way to do the configuration automatically and keep a consistent schelling game between testnet and mainnet (eg: a slow disk / IO on a node may result in a working schelling game on testnet due to the longer block time, but won't on gnosis chain due to the shorter block time).

PR #112, opened 2023-02-01
feat: postage stamp index checkings

nugaon:
Added 3 postage stamp index related checks

### DIFF
--- a/src/Redistribution.sol
+++ b/src/Redistribution.sol
@@ -74,7 +74,7 @@ contract Redistribution is AccessControl, Pausable {
         // proveSegmentIndex3 know, is equal _proveSegmentIndex2;
         // chunkSpan2 is equal to chunkSpan (as the data is the same)
 
-        address signer;
+        // address signer; it is provided by the postage stamp contract
         bytes signature;
         bytes32 chunkAddr;
         bytes32 postageId;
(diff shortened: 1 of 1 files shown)


n00b21337 (commented):
Currently, this is failing as you have

const tx2 = await r_node_2.claim();  on line 868  which you made a new function and expects 3 params


        ChunkInclusionProof calldata entryProof1,
        ChunkInclusionProof calldata entryProof2,
        ChunkInclusionProof calldata entryProofLast

n00b21337 (commented):
```
export function proximity(hexA: string, hexB: string): number {
  const one = hexToBytes(hexA);
  const other = hexToBytes(hexB);
```

As this code is used only in tests and I suppose it will be used only there, should be in **tests/util** directory ?

n00b21337:
> Currently, this is failing as you have
> 
> `const tx2 = await r_node_2.claim();` on line 868 of tests and as you made a new claim function it expects 3 params
> 
> ```
>     ChunkInclusionProof calldata entryProof1,
>     ChunkInclusionProof calldata entryProof2,
>     ChunkInclusionProof calldata entryProofLast
> ```

also the Stats.test.ts is failing for the same reason

await r_node.claim(); on line 106 should have 3 params

nugaon:
> await r_node.claim(); on line 106 should have 3 params

I'm aware of this, and I'm going to fix it soon. 
Thinking about the possibilities, maybe I change each cases to the `node_5` claim since I have the test fixture for that only.

n00b21337 (commented):
verify.ts shouldn't be removed as its used for verifying  SCs. bringing it back

PR #115, opened 2023-02-14
fix: how to properly import other contracts functionality with interface

n00b21337:
If we take a look at the contract deployed here
https://gnosisscan.io/address/0x30d155478ef27ab32a1d578be7b84bc5988af381#code

you can see it has 1760 lines of code because it imports the whole Stamps contract. 
The proper pattern to interact with other SC is to use the **interface** and declared methods you need to use. For example, with those changes, the Oracle contract when deployed will have 130 lines of code and does exact same thing.
(also this produces smaller bytecode and takes less space on the blockchain)
Passes all the tests and works the same. 
I know Gnosis is not so expensive for deployments and this would be much more effective on L1 but this is just a better, cleaner, and more professional code.


n00b21337:
I am not informed how CD/CI process works and still waiting for some answers, so apologies if this is not correct. I just wanted to make a proposal about how things could be improved.

n00b21337:
I added interface use for all the contracts that can use them. Locally deployment and tests go through. Need to see what is problem with CI.

mrekucci (commented):
Nice cleanup. I would recommend avoiding the `I` prefix in interfaces. My understanding is that there is no such convention in Solidity, and the user doesn't need to know that he is dealing with an interface: https://docs.soliditylang.org/en/latest/style-guide.html#naming-conventions.

n00b21337:
> Nice cleanup. I would recommend avoiding the `I` prefix in interfaces. My understanding is that there is no such convention in Solidity, and the user doesn't need to know that he is dealing with an interface: https://docs.soliditylang.org/en/latest/style-guide.html#naming-conventions.

yes it's not required but it is considered good practice to prefix your interfaces with I. You will often see defined interfaces like IERC20 etc all having the prefix
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/IERC20.sol

Here are some more links and tutorials that also recommend it
https://coinsbench.com/solidity-101-introduction-to-interfaces-in-solidity-8b0bd64eda38
https://medium.com/coinmonks/solidity-tutorial-all-about-interfaces-f547d2869499

Also, there is a practical reason, removing I and naming them the same as contracts will confuse hardhat with the current setup and will throw "cannot find artifact" for deployment scripts.

mrekucci:
> yes it's not required but it is considered good practice to prefix your interfaces with I. You will often see defined interfaces like IERC20 etc all having the prefix https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/IERC20.sol

If a project uses it, it doesn't necessarily mean it's good practice, one can find a number of [counterexamples](https://www.google.com/search?q=prefixing+interfaces+with+I&oq=prefixing++interfaces+with+I&aqs=chrome..69i57j0i390l2j69i60.6269j0j4&sourceid=chrome&ie=UTF-8) with solid justification.

> Here are some more links and tutorials that also recommend it https://coinsbench.com/solidity-101-introduction-to-interfaces-in-solidity-8b0bd64eda38 https://medium.com/coinmonks/solidity-tutorial-all-about-interfaces-f547d2869499

Both of these sources merely state that the convention is to name interfaces beginning with `I` without any justification or reference to the source.

> Also, there is a practical reason, removing I and naming them the same as contracts will confuse hardhat with the current setup and will throw "cannot find artifact" for deployment scripts.

If that's the case, then I would recommend making a reasonable fix.

The interface prefix `I` comes mainly from languages such as Java and C#. In some other languages, the `I` prefix is also used to denote the implementation of an interface, not the interface itself. My point is that if we want to follow this convention, there should be a sound justification for it.

mrekucci (approved):

n00b21337:
> > yes it's not required but it is considered good practice to prefix your interfaces with I. You will often see defined interfaces like IERC20 etc all having the prefix https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/IERC20.sol
> 
> If a project uses it, it doesn't necessarily mean it's good practice, one can find a number of [counterexamples](https://www.google.com/search?q=prefixing+interfaces+with+I&oq=prefixing++interfaces+with+I&aqs=chrome..69i57j0i390l2j69i60.6269j0j4&sourceid=chrome&ie=UTF-8) with solid justification.
> 
> > Here are some more links and tutorials that also recommend it https://coinsbench.com/solidity-101-introduction-to-interfaces-in-solidity-8b0bd64eda38 https://medium.com/coinmonks/solidity-tutorial-all-about-interfaces-f547d2869499
> 
> Both of these sources merely state that the convention is to name interfaces beginning with `I` without any justification or reference to the source.
> 
> > Also, there is a practical reason, removing I and naming them the same as contracts will confuse hardhat with the current setup and will throw "cannot find artifact" for deployment scripts.
> 
> If that's the case, then I would recommend making a reasonable fix.
> 
> The interface prefix `I` comes mainly from languages such as Java and C#. In some other languages, the `I` prefix is also used to denote the implementation of an interface, not the interface itself. My point is that if we want to follow this convention, there should be a sound justification for it.

Thank you for the first review :muscle: :smile: 

About interfaces and naming, I did read what you linked. Would like to state, that I just showed a few examples, defining interfaces with I is actually standard in solidity.
https://github.com/aave/aave-v3-core/tree/master/contracts/interfaces
https://github.com/Uniswap/v3-core/tree/main/contracts/interfaces
Some more prominent examples here, actually I don't know if any projects don't do it like that. Hmm yes, there are some but they still state it in name, as ChainLink does with "AggregatorV3Interface" for example.
The bottom line here is the strong convention to always name the interface in this ecosystem of contracts, tools, and developers.

About tooling, yes, there is a problem with namespacing in cases where there is for example interface with the name PostageStamp and there is a contract named PostageStamp, hardhat gets broken but it can be fixed. With the use of fully qualified naming in deployments and tests it does work and we can have the same name.

I do agree that it looks nicer without the I prefix but then again solidity tries to be very exact because of possible and costly errors so I do understand that this standard took off just to be more clear.

src/Redistribution.sol
```diff
@@ -191,7 +214,7 @@ contract Redistribution is AccessControl, Pausable {
      */
     function currentPhaseReveal() public view returns (bool) {
         uint256 number = block.number % roundLength;
-        if ( number >= roundLength / 4 && number < roundLength / 2 ) {
+        if (number >= roundLength / 4 && number < roundLength / 2) {
```
significance, line 217: let's make sure commits do not have formatting changes to improve readability 🤓

significance:
can we remove the package-lock.json file from version control please?

significance (approved):
aside from couple of minor complaints, looks good to me!

src/Redistribution.sol
n00b21337: Sorry for that. Usually, I do formatting changes in one commit and actual changes in another so it's easier to track what is what.

n00b21337:
> can we remove the package-lock.json file from version control, please?

I actually use both yarn and npm but more often npm. That being said if we are using yarn for CI then seems it's better to use that and not update the lock file of npm but for yarn.  I will remove it.

Issue #87, opened 2022-12-12
Implement London transaction


mrekucci:
Add a smart contract deployment script that will use the new EIP-1559 (London) transaction.

n00b21337:
Connecting current PR
https://github.com/ethersphere/storage-incentives/pull/92/commits

n00b21337:
![Screenshot from 2023-03-09 08-30-57.png](https://images.zenhubusercontent.com/640746a421238e2a06909e41/902ff6c0-3910-4905-b574-8729b583712b)

To make this task it is needed to update libraries used with hardhat as most of them are outdated as seen from the image above and the current state. I would do this in another branch as the previous guy did this but also added some others and not sure if he updated all properly.  Also, the task should be split into 2 things. The whole scripts deployment is not needed. We have been using it for tests  ethers-deploy library and this library should also be used for actual deployment, that is what it is made for, there is no point in having extra scripts/deployer  script and running it. Please advise If I am missing something here and why it was used.

n00b21337:
This is done and deployed.

PR #119, opened 2023-03-09
feat: Update libraries to get EIP1559 trx

n00b21337:
This is PR that is connected to the task here 
https://github.com/ethersphere/storage-incentives/issues/87

as said in that issue. I updated libraries which in turn use an EIP-1559 type of trx as default.
With that, I also made a series of smaller fixes to repo/env.  All are commented with commits.

The second part of this task would be to eradicate deployer.ts and just use scripts that are in the deploy folder which is the intended way to use [ethers-deploy ](https://github.com/wighawag/hardhat-deploy).


significance (approved):
lgtm!

hardhat.config.ts
```diff
@@ -103,6 +104,16 @@ const config: HardhatUserConfig = {
           privateKey: '0xc1a7ca0bc39058d1fb6e331f8a6a3b65e81f171cfeddeec6cd0f32271496f45c',
           balance: '10000000000000000000000',
         },
+        // other_1
```
mrekucci, line 107: Why are these new accounts needed?

mrekucci (approved):

PR #120, opened 2023-03-10
fix: Pot change to view

n00b21337:
Old code for checking total


n00b21337:
wanted to rework commits so its easier to review, opened another branch

PR #121, opened 2023-03-10
fix: Improve rewards function

n00b21337:
This is a small fix for the **totalPot** function. The problem is that function is at this moment WRITE type which means you can't get current rewards without making a transaction. I made it into READ mode, which means users or apps can just call it and get value, expireLimited is called where it is supposed to be. Functionality is the same, just UX is better.
Also, I proposed a naming change to "currentReward" as if you see the code it's actually not total pot but could also be what is there in balance, so the reward could be lower than pot. Also totalpot value is fetchable by default getter because pot is a public variable.

The first commit is code formatting by prettier so just check second commit.

### DIFF
--- a/src/PostageStamp.sol
+++ b/src/PostageStamp.sol
@@ -447,10 +447,9 @@ contract PostageStamp is AccessControl, Pausable {
     }
 
     /**
-     * @notice The current pot.
+     * @notice The current reward value
      */
-    function totalPot() public returns (uint256) {
-        expireLimited(type(uint256).max);
+    function currentReward() public view returns (uint256) {
         uint256 balance = ERC20(bzzToken).balanceOf(address(this));
         return pot < balance ? pot : balance;
     }
@@ -462,7 +461,8 @@ contract PostageStamp is AccessControl, Pausable {
 
     function withdraw(address beneficiary) external {
         require(hasRole(REDISTRIBUTOR_ROLE, msg.sender), "only redistributor can withdraw from the contract");
-        uint256 totalAmount = totalPot();
+        expireLimited(type(uint256).max);
+        uint256 totalAmount = currentReward();
         require(ERC20(bzzToken).transfer(beneficiary, totalAmount), "failed transfer");
 
         emit PotWithdrawn(beneficiary, totalAmount);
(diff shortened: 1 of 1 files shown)


significance:
maybe splitting hairs here but technically the totalPot/currentReward is not accurate/up to date without calling expireLimited. with this pattern, one could check the totalPot/currentReward, then withdraw to find out the actual value is larger.

could you explain a bit more why you think this is an improvement to UX? perhaps this is something to do with how Bee is using the api?

in my mind neither solution is great and both suffer from poor UX so i don't really have any  preference! 

similarly am ok with the name change if you have a strong preference, although for me the 'pot' is what can be 'won' in that 'round' so it does make sense

n00b21337:
> maybe splitting hairs here but technically the totalPot/currentReward is not accurate/up to date without calling expireLimited. with this pattern, one could check the totalPot/currentReward, then withdraw to find out the actual value is larger.
> 
> could you explain a bit more why you think this is an improvement to UX? perhaps this is something to do with how Bee is using the api?
> 
> in my mind neither solution is great and both suffer from poor UX so i don't really have any preference!
> 
> similarly am ok with the name change if you have a strong preference, although for me the 'pot' is what can be 'won' in that 'round' so it does make sense

That is good point.  The idea is that public functions are usually something you can either use to read state like in etherscan and/or you call it by other contracts. Basically this one needs gas to run because of expireLimited() that changes state in it.  So it wouldn't be a good UX for users that want to check state of pot or for bee node app also if it uses it somewhere as now it needs to make trx and pay for gas. That part is not good.

Your point is valid, if they want to check state they might not get the most recent one, then again this could happen even with current setup as values change each block, so there could be more expired stamps between checking state and wdraw, depending on how much they wait between too. Not sure about this in the end, still maybe looks better to me that pot/reward can be checked without trx. Also if Bee is not using it, then there is no need for it to be public, it can be internal so we save on gas (yeah we might say its cheap but its good to optimize for gas all the time)

About the naming, not against the pot as name, what I find the slight issue is this
"return pot < balance ? pot : balance;"

which means it could be sometimes its not really the pot you get, so just trying to align with logic of that.

p.s.
There is pot getter as I said but it might not be accurate also if we take into account this logic
"return pot < balance ? pot : balance;"

I know we are a bit bikeshedding now, i just like to get tiny details right also :) this is a detail and really not important in the end.

n00b21337:
I will close this one. Not sure if there is any improvement in what I proposed. There is default pot getter users can get which in most cases would be the same as the thing I did.

PR #123, opened 2023-03-16
refactor: Redistribution

n00b21337:
This task is mainly about cleaning the code to make it more readable and conscience. 
- Purpose of that is that it helps to read and understand code which in the end helps make fewer errors which is paramount when working with solidity
- I made pretty clear changes per each commit to what is going on so I suggest following it like that for CR
- Main change was focused on the claim function making it more readable by changing complex if/else structures which are really not a good way to code https://softwareengineering.stackexchange.com/questions/206816/clarification-of-avoid-if-else-advice
- Synched up the claim and isWinner function to have the same code
- I also changed some vars grouping 
- In the end, improved Stats test group for a better setup
- Also helped me get more accustomed with codebase

I also wanted to make claim/isWinner more reusable and reuse code as they share most of the code. But there is a big problem as one changes state and the other doesn't, so I couldn't find a proper way to refactor this unless I change the type of isWinner to write.

### DIFF
--- a/src/Redistribution.sol
+++ b/src/Redistribution.sol
@@ -598,33 +598,43 @@ contract Redistribution is AccessControl, Pausable {
 
         for (uint256 i = 0; i < commitsArrayLength; i++) {
             revIndex = currentCommits[i].revealIndex;
-            if (currentCommits[i].revealed) {
+
+            // Select winner with valid truth
+            if (
+                currentCommits[i].revealed &&
+                truthRevealedHash == currentReveals[revIndex].hash &&
+                truthRevealedDepth == currentReveals[revIndex].depth
+            ) {
+                currentWinnerSelectionSum += currentReveals[revIndex].stakeDensity;
+                randomNumber = keccak256(abi.encodePacked(winnerSelectionAnchor, k));
+
+                randomNumberTrunc = uint256(randomNumber & MaxH);
+
                 if (
-                    truthRevealedHash == currentReveals[revIndex].hash &&
-                    truthRevealedDepth == currentReveals[revIndex].depth
+                    randomNumberTrunc * currentWinnerSelectionSum <
+                    currentReveals[revIndex].stakeDensity * (uint256(MaxH) + 1)
                 ) {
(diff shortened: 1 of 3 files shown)


src/Redistribution.sol
```diff
@@ -162,14 +162,9 @@ contract Redistribution is AccessControl, Pausable {
...
-     * @dev Emits the number of commits being processed by the claim phase.
+     * @dev Emits the number of commits and reveals being processed by the claim phase.
      */
-    event CountCommits(uint256 _count);
-
-    /**
-     * @dev Emits the number of reveals being processed by the claim phase.
-     */
-    event CountReveals(uint256 _count);
+    event CountCommitsReveals(uint256 _count, uint256 _count);
```
significance: we should avoid changing events unless we need to in order not to break compatibility

n00b21337: makes sense, will put it as it was.

mrekucci (approved):
Nice.

test/Redistribution.test.ts
```diff
@@ -827,4 +827,4 @@ describe('Redistribution', function () {
       });
     });
   });
-});
+});
```
nugaon, line 830: insert a new line here pls.

src/Redistribution.sol
```diff
@@ -618,56 +566,88 @@ contract Redistribution is AccessControl, Pausable {
...
+                    truthRevealedDepth != currentReveals[revIndex].depth)
+            ) {
+                Stakes.freezeDeposit(
+                    currentReveals[revIndex].overlay,
+                    penaltyMultiplierDisagreement * roundLength * uint256(2 ** truthRevealedDepth)
+                );
+            }
+
+            // Slash deposits if revealed is false
+            if (!currentCommits[i].revealed) {
```
nugaon, line 632: the original `if..else` seems a bit more optimal and readable for me.

nugaon (approved):
LGTM!

test/Redistribution.test.ts
n00b21337: adding it in the next PR

PR #124, opened 2023-03-21
fix: Refactor postageStamp

n00b21337:
Just some code refactoring.
First commit is prettier reformat, second is changes.
Meaningful variable name and grouping of same IF condition


n00b21337:
had and error in logic, closing

PR #127, opened 2023-03-27
Streamline deployment

n00b21337:


n00b21337:
Also a note on codegen/generate_src.sh  file. Not sure if this  is correct but I suppose we don't need it, mainet_deployed.json as well as tesnet_deployed.json can be fully generated from TS code. We have saved ABI and the rest of the data for deployed BZZ and gBZZ which I put to reside in helper-hardhat-config.ts 
If I am missing some logic here that is important please inform me, my logic is that this rarely changes, it can be stored as vars once deployed and reused in future deployments.  I could also retrieve it from chain and etherscan with saved deployed address and trx hash but seems to me obsolete in this case to make those calls.

deploy/000_deploy_postage.ts
```diff
@@ -1,29 +1,36 @@
 import { HardhatRuntimeEnvironment } from 'hardhat/types';
 import { DeployFunction } from 'hardhat-deploy/types';
+import { networkConfig, developmentChains } from '../helper-hardhat-config';
```
nugaon, line 3: not used `HardhatRuntimeEnvironment` and `developmentChains` imports

nugaon (commented):
I superficially checked the old `deployer.ts` and I would keep its console.log messages.
The hardhat scripts multiple times change the same role that may need a refactor.
there are multiple unused imports and other stuff that I commented below.

deploy/000_deploy_postage.ts
```diff
@@ -1,29 +1,36 @@
...
 
+  const args = [token.address, 16];
+
   await deploy('PostageStamp', {
     from: deployer,
-    args: [token.address, 16],
+    args: args,
     log: true,
+    // we need to wait if on a live network so we can verify properly
+    waitConfirmations: networkConfig[network.name].blockConfirmations || 1,
```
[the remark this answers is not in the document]
n00b21337: this is better, will apply, thnx

deploy/001_deploy_oracle.ts
```diff
@@ -1,19 +1,25 @@
 import { HardhatRuntimeEnvironment } from 'hardhat/types';
 import { DeployFunction } from 'hardhat-deploy/types';
+import { networkConfig, developmentChains } from '../helper-hardhat-config';
+import verify from '../utils/verify';
```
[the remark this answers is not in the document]
n00b21337: was solved even before the comment, should have let you know to update the local git. sry

deploy/002_deploy_staking.ts
```diff
@@ -1,21 +1,30 @@
...
-  const { deploy, get, read, execute } = deployments;
+const func: DeployFunction = async function ({ deployments, getNamedAccounts, network }) {
+  const { deploy, get, read, execute, log } = deployments;
   const { deployer, pauser } = await getNamedAccounts();
 
-  const networkID = 0; //test network
+  // Overlays in tests are hardcoded with 0 ID so we need to use it for testing
+  let networkID = 0;
+  if (!developmentChains.includes(network.name)) {
+    networkID = network.config.chainId!;
```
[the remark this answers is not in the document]
n00b21337: better, applied, thnx

hardhat.config.ts
```diff
@@ -119,17 +119,16 @@ const config: HardhatUserConfig = {
...
     localhost: {
       url: 'http://localhost:8545',
-      accounts,
-      chainId: 1,
+      chainId: 31337,
     },
     testnet: {
       url: 'https://goerli.infura.io/v3/' + infuraToken,
-      accounts,
+      accounts: [], // we need to add private keys here
```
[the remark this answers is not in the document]
n00b21337: oh yeah, was doing that in follow-up PR but agree, this was suboptimal approach.

deploy/002_deploy_staking.ts
```diff
@@ -1,21 +1,30 @@
...
+    waitConfirmations: networkConfig[network.name].blockConfirmations || 1,
   });
 
   const pauserRole = await read('StakeRegistry', 'PAUSER_ROLE');
   await execute('StakeRegistry', { from: deployer }, 'grantRole', pauserRole, pauser);
+  log('----------------------------------------------------');
 };
 
 export default func;
+func.tags = ['all', 'staking', 'contracts'];
```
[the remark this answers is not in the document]
n00b21337: These are deployment tags, so basically If I want to run the deployment of all scripts I would write
hh deploy --tags "all"
If I want to deploy just one or just verify I would write
hh deploy --tags "verify"

I usually add them just as good practice as sometimes you might need them but in our case, it's not so useful as each step is connected to others so it won't help in the current setup unless they are run altogether.

deploy/003_deploy_redistribution.ts
```diff
@@ -22,7 +25,9 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     { from: deployer },
     'grantRole',
     redistributorRoleStakeRegistry,
-    (await get('Redistribution')).address
+    (
+      await get('Redistribution')
+    ).address
```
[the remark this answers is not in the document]
n00b21337: hmm not sure what should I do here. I have prettier and this is how it autoformats to me. Maybe we have some different settings in them?

hardhat.config.ts
```diff
@@ -119,17 +119,16 @@ const config: HardhatUserConfig = {
     },
     localhost: {
       url: 'http://localhost:8545',
-      accounts,
-      chainId: 1,
+      chainId: 31337,
```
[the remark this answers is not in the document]
n00b21337: chainId 1 is usually used for Ethereum and default hardhat ID is 31337 so I just set it at that, think it a good practice to use defaults when it's not a problem 
https://hardhat.org/hardhat-network/docs/reference#chainid

Also before there was a problem with metamask and you couldn't use it without setting another ID which was "1337" but from what I see in practice this is no more a problem at least for metamask. Then again I like to set it in code so it is seen which port it is for that network.

https://hardhat.org/hardhat-network/docs/metamask-issue
https://github.com/MetaMask/metamask-extension/issues/10290

n00b21337:
> I superficially checked the old `deployer.ts` and I would keep its console.log messages. The hardhat scripts multiple times change the same role that may need a refactor. there are multiple unused imports and other stuff that I commented below.

The reason for using **log** instead of **console.log** is we are using this library
https://github.com/wighawag/hardhat-deploy
where is a common way to log out info with that method.

About the roles, please give more info on that, I would gladly refactor what is needed.

helper-hardhat-config.ts
```diff
@@ -0,0 +1,881 @@
...
+    blockConfirmations: 6,
+  },
+  mainnet: {
+    blockConfirmations: 6,
+  },
+};
+
+export const developmentChains = ['hardhat', 'localhost'];
+
+export const deployedBzzData = {
```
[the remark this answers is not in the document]
nugaon: yeah, that one. it is really similar meta file that the deployment produces, isn't it?
if so, it could be `bzz-token-meta.json` or smth.

deploy/003_deploy_redistribution.ts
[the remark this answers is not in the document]
nugaon: hm... indeed, now I don't have prettier problems either. Maybe there was some temporary bug in my IDE

deploy/005_deploy_local_data.ts
```diff
@@ -0,0 +1,112 @@
...
+    network.name == 'mainnet'
+      ? (deployedData['contracts']['bzzToken'] = deployedBzzData.mainnet)
+      : (deployedData['contracts']['bzzToken'] = deployedBzzData.testnet);
+  } else {
+    // Token data for dev chains
+    deployedData['contracts']['bzzToken']['abi'] = tokenContract.abi;
+    deployedData['contracts']['bzzToken']['bytecode'] = tokenContract.bytecode ? tokenContract.bytecode : '';
+    deployedData['contracts']['bzzToken']['address'] = tokenContract.address;
+    deployedData['contracts']['bzzToken']['block'] =
+      tokenContract.receipt && tokenContract.receipt.blockNumber !== undefined ? tokenContract.receipt.blockNumber : 0;
```
nugaon, line 69:
```suggestion
      tokenContract.receipt && tokenContract.receipt.blockNumber ? tokenContract.receipt.blockNumber : 0;
```

nugaon: same at other places as well

hardhat.config.ts
```diff
@@ -1,15 +1,16 @@
...
 import 'hardhat-tracer';
 import '@nomiclabs/hardhat-etherscan';
 
-const infuraToken = process.env.INFURA_TOKEN === undefined ? 'undefined' : process.env.INFURA_TOKEN;
-if (infuraToken === 'undefined') {
-  console.log('Please set your INFURA_TOKEN in a .env file');
-}
+// Set Private RPCs if added, otherwise use Public that are hardcoded in this config
+const PRIVATE_RPC_MAINNET =
+  process.env.PRIVATE_RPC_MAINNET === undefined ? 'undefined' : process.env.PRIVATE_RPC_MAINNET;
```
nugaon, line 11:
```suggestion
  !process.env.PRIVATE_RPC_MAINNET ? 'undefined' : process.env.PRIVATE_RPC_MAINNET;
```

nugaon (commented):
The `TestToken` related changes can be sorted out in a different PR.

For roles, I meant e.g. `PRICE_ORACLE_ROLE` is set two times at postage and oracle contract deployments. This is useful when only one of the scripts is executed, but takes an additional unnecessary cost when the other script has run before. This is just a slight optimization in the code, it can happen later in a different PR.

deploy/002_deploy_staking.ts
```diff
@@ -1,21 +1,28 @@
...
   const { deployer, pauser } = await getNamedAccounts();
 
-  const networkID = 0; //test network
+  // Overlays in tests are hardcoded with 0 ID so we need to use it for testing
+  let networkID = 0;
+  if (!developmentChains.includes(network.name) && network.config.chainId) {
+    networkID = network.config.chainId;
+  }
 
+  const args = [(await get('TestToken')).address, networkID];
```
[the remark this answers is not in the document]
n00b21337: thank you, I missed this important part. fixing now. added also that part for postage stamp now so we don't wait for second PR

README.md
```diff
@@ -96,49 +104,56 @@ To get started with this project, follow these steps:
 ## Run
 
 ### [Tests](./test)
+
 - Unit Tests
-  - Run `yarn run test` to run all the tests.
-  - Run `yarn run test:coverage` to see the coverage of smart contracts.
+  - Run `hardhat test` to run all the tests.
```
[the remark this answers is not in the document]
n00b21337: yes. I use https://hardhat.org/hardhat-runner/docs/guides/command-line-completion  
so i just write "hh test" and forgot that its usually not just that, will add yarn or npx

.eslintrc.js
```diff
@@ -11,5 +11,6 @@ module.exports = {
   rules: {
     // Place to specify ESLint rules. Can be used to overwrite rules specified from the extended configs
     // e.g. "@typescript-eslint/explicit-function-return-type": "off",
+    '@typescript-eslint/no-explicit-any': 'off',
```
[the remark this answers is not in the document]
n00b21337: :D  I thought you might say that but you did say to "change the rule". Ok I suppose using any is not the best practice and should be avoided, removing this lint skip and adding unknown and void/type check.

hardhat.config.ts
```diff
@@ -6,7 +6,7 @@ import 'hardhat-deploy-ethers';
 import 'hardhat-tracer';
 import '@nomiclabs/hardhat-etherscan';
 
-const infuraToken = process.env.INFURA_TOKEN === undefined ? 'undefined' : process.env.INFURA_TOKEN;
+export const infuraToken = process.env.INFURA_TOKEN === undefined ? 'undefined' : process.env.INFURA_TOKEN;
```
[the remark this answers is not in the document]
nugaon: this was actually my bad, I left out the exclamation mark from the beginning, sorry for that.

deploy/000_deploy_postage.ts
```diff
@@ -1,29 +1,48 @@
...
+  let token = null;
+  if (developmentChains.includes(network.name)) {
+    token = await deploy('TestToken', {
+      from: deployer,
+      args: [],
+      log: true,
+    });
+  }
+
+  if (network.name == 'testnet') {
```
mrekucci, line 17: This can be simplified to:
```
if (network.name == 'mainnet' || network.name == 'testnet') {
    token = await ethers.getContractAt(deployedBzzData[network.name].abi, deployedBzzData[network.name].address);
}
```
  
Then the if below can be removed.

hardhat.config.ts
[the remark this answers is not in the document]
n00b21337: not a problem, this is already implemented on improved version of this that is not tied to infura but can use any private RPC...
think you already also reviewed that

`const PRIVATE_RPC_TESTNET = !process.env.PRIVATE_RPC_TESTNET ? 'undefined' : process.env.PRIVATE_RPC_TESTNET;`

deploy/005_deploy_local_data.ts
```diff
@@ -0,0 +1,112 @@
...
+  // Insert already deployed data if it is mainnet or testnet
+  if (!developmentChains.includes(network.name)) {
+    network.name == 'mainnet'
+      ? (deployedData['contracts']['bzzToken'] = deployedBzzData.mainnet)
+      : (deployedData['contracts']['bzzToken'] = deployedBzzData.testnet);
+  } else {
+    // Token data for dev chains
+    deployedData['contracts']['bzzToken']['abi'] = tokenContract.abi;
+    deployedData['contracts']['bzzToken']['bytecode'] = tokenContract.bytecode ? tokenContract.bytecode : '';
+    deployedData['contracts']['bzzToken']['address'] = tokenContract.address;
```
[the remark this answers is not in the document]
n00b21337: Let me explain what is happening there.

What you said is what it is doing. Code in IF statement before that uses deployed data on testnet and mainet as we don't deploy new contracts for BZZ and we used previously deployed ones, but locally we don't have something we use from before, we use the currently deployed token in step 0, so basically this is what that code does. 
Dev chains lose state after you close them so you can't refer to anything from before even if you wanted to use something previously deployed so it wouldn't make sense to put here anything other then what it is now

deploy/000_deploy_postage.ts
n00b21337: lol, obviously. did some refactoring and didnt even realize I ended with same 2 lines. Fixing this, thnx

deploy/002_deploy_staking.ts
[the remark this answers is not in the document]
n00b21337: yes, that would work. Its a pattern I use from a good teacher of mine :) Patrick Collins 
https://github.com/PatrickAlphaC/hardhat-fund-me-fcc/tree/main/deploy

So the idea is to put "all" for the main procedure, but if you add some extra steps, then you just put specific tag. Kind of "main" thread or main deployment process. But also to have option to have other specific steps that are not into that main. Seems to me "main" would be better name, but used "all" for now.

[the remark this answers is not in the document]
n00b21337: Also a note on what I said before. locally for hardhat this tags are not useful as hh resets state as soon as deployment is done. But if you run hh localhost or when you deploy to test or main, then deployments are saved in the deployments folder with full info. Which is great as then you can just use each of the step. For example we can just redeploy "redistribution" contract with this same script on top of all what was deployed already, so this process becomes very useful and tags are good to have.

n00b21337:
> The `TestToken` related changes can be sorted out in a different PR.
> 
> For roles, I meant e.g. `PRICE_ORACLE_ROLE` is set two times at postage and oracle contract deployments. This is useful when only one of the scripts is executed, but takes an additional unnecessary cost when the other script has run before. This is just a slight optimization in the code, it can happen later in a different PR.

ohh yeah, that is excellent remark, was thinking about that, but slipped my mind. Roles should be more streamlined and this has some obsolete code.

n00b21337:
I am refactoring ROLES setters when deploying, tests are falling at this moment, so don't CR tests until I improve that.

test/PostageStamp.test.ts
```diff
@@ -52,6 +53,8 @@ describe('PostageStamp', function () {
   describe('when deploying contract', function () {
     beforeEach(async function () {
       await deployments.fixture();
+      const priceOracleRole = await read('PostageStamp', 'PRICE_ORACLE_ROLE');
```
significance, line 56: why are these grant roles needed here? is it not handled in `004_deploy_roles_postage.ts`?

deploy/002_deploy_staking.ts
```diff
@@ -1,21 +1,39 @@
...
+  const args = [token.address, networkID];
   await deploy('StakeRegistry', {
     from: deployer,
-    args: [(await get('TestToken')).address, networkID],
+    args: args,
     log: true,
+    waitConfirmations: networkConfig[network.name]?.blockConfirmations || 1,
   });
 
-  const pauserRole = await read('StakeRegistry', 'PAUSER_ROLE');
```
significance, line 17: why is this removed here?

n00b21337: Previous deployment scripts were used just for local setup. Which meant we were using all those local accounts set in hardhat settings. There are also pauser and oracle accounts.
So they are just local EOA addresses, which we cant use on public chains.  If we want to use some additional specific account to be pauser on staking I need to know which address would that be but I assume we dont need that. And default pauser is set already in constructor of staking contrat.

` _setupRole(PAUSER_ROLE, msg.sender);`

test/PostageStamp.test.ts
n00b21337: Yes, those 2 lines for that test are actually not needed, and have just removed them.

n00b21337: Also for the whole test take into account that some roles are set to contract addresses, so when you test something you got simulate as it was called from the contract.

deploy/002_deploy_staking.ts
significance: understood, thanks @n00b21337

significance (approved):

deploy/000_deploy_postage.ts
```diff
@@ -1,29 +1,38 @@
-import { HardhatRuntimeEnvironment } from 'hardhat/types';
 import { DeployFunction } from 'hardhat-deploy/types';
+import { networkConfig, developmentChains, deployedBzzData } from '../helper-hardhat-config';
 
-const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
-  const { deployments, getNamedAccounts } = hre;
-  const { deploy, execute, read } = deployments;
+const func: DeployFunction = async function ({ deployments, getNamedAccounts, network, ethers }) {
+  const { deploy, log } = deployments;
+  const { deployer } = await getNamedAccounts();
```
nugaon, line 6: it is a bit ugly here the `deployer` can be retrieved from `getNamedAccounts`...

deploy/000_deploy_postage.ts
```diff
@@ -1,29 +1,38 @@
...
 
-  const { deployer, oracle, redistributor } = await getNamedAccounts();
+  let token = null;
```
nugaon, line 8:
```suggestion
  let token: Contract | null = null;
```
use types whenever you can and don't use the same variable for different types when it is not justified.

deploy/002_deploy_staking.ts
```diff
@@ -1,21 +1,39 @@
...
+    token = await get('TestToken');
+  }
+
+  if (network.name == 'mainnet' || network.name == 'testnet') {
+    token = await ethers.getContractAt(deployedBzzData[network.name].abi, deployedBzzData[network.name].address);
+  }
+
+  if (token == null) {
+    throw new Error(`Unsupported network: ${network.name}`);
+  }
```
nugaon, line 25: could be a utility function returning `token`

deploy/007_deploy_verify.ts
```diff
@@ -0,0 +1,49 @@
+import { DeployFunction } from 'hardhat-deploy/types';
+import { developmentChains, deployedBzzData } from '../helper-hardhat-config';
```
nugaon, line 2: not used variable

deploy/007_deploy_verify.ts
```diff
@@ -0,0 +1,49 @@
...
+import verify from '../utils/verify';
+
+const func: DeployFunction = async function ({ deployments, network, ethers }) {
+  const { log, get } = deployments;
+
+
```
nugaon, line 8: unnecessary breakline

nugaon (approved):
LGTM!

deploy/000_deploy_postage.ts
n00b21337: Not sure what is suggestion here. Can you reiterate?

nugaon: at first glance I thought that is a deployer instance and actually I wanted to delete the comment after.

PR #128, opened 2023-03-30
New gbzz

n00b21337:
We add a new deployment step that is making test token on testnet.
Token is deployed with EOA wallet but switched to multisig right after deployment.
All the tokens are minted to this wallet which can then distribute around.

(apologies for the commit history, did checkout  on the basis of the feature branch)


mrekucci (approved):
Nice.

PR #132, opened 2023-04-03
fix: add new redis events

n00b21337:
Ok, I did apply all the changes we have in the current master of redis as they are mostly cosmetic. 
All tests pass with the current version, old deployment scripts also work on this new version.

Important changes for this task are
- we already have round emitted for commit here so it's not needed to add
https://github.com/ethersphere/storage-incentives/blob/83d585bf0f6e1a6fb86b4266128d5392e5875e64/src/Redistribution.sol#L303
- line 398 https://github.com/ethersphere/storage-incentives/blob/83d585bf0f6e1a6fb86b4266128d5392e5875e64/src/Redistribution.sol#L398
where we emit current round and anchor
- line 669 https://github.com/ethersphere/storage-incentives/blob/83d585bf0f6e1a6fb86b4266128d5392e5875e64/src/Redistribution.sol#L669
where we add the current round to the winner event, so we can associate the winner with the winning round and/or just fetch the current round for this claim 
- line 670 https://github.com/ethersphere/storage-incentives/blob/83d585bf0f6e1a6fb86b4266128d5392e5875e64/src/Redistribution.sol#L670
where we now have validChunkCount added as an event, also added pot size, which could be useful but can remove

Additional note, there is also round number emmited for reveal in each reveal event
https://github.com/ethersphere/storage-incentives/blob/83d585bf0f6e1a6fb86b4266128d5392e5875e64/src/Redistribution.sol#L427
but as we need to emit anchor, it seemed ok to emit also round number there.

### DIFF
--- a/src/Redistribution.sol
+++ b/src/Redistribution.sol
@@ -2,9 +2,35 @@
 pragma solidity ^0.8.1;
 import "@openzeppelin/contracts/access/AccessControl.sol";
 import "@openzeppelin/contracts/security/Pausable.sol";
-import "./PostageStamp.sol";
-import "./PriceOracle.sol";
-import "./Staking.sol";
+
+/**
+ * Implement interfaces to PostageStamp contract, PriceOracle contract and Staking contract.
+ * For PostageStmap we currently use "withdraw" to withdraw funds from Pot
+ * For PriceOracle we use "adjustPrice" to change price of PostageStamps
+ * For Staking contract we use "lastUpdatedBlockNumberOfOverlay, freezeDeposit, ownerOfOverlay, stakeOfOverlay"
+ */
+
+interface IPostageStamp {
+    function withdraw(address beneficiary) external;
+
+    function validChunkCount() external view returns (uint256);
+
+    function pot() external view returns (uint256);
+}
+
+interface IPriceOracle {
+    function adjustPrice(uint256 redundancy) external;
+}
+
+interface IStakeRegistry {
+    function lastUpdatedBlockNumberOfOverlay(bytes32 overlay) external view returns (uint256);
+
+    function freezeDeposit(bytes32 overlay, uint256 time) external;
+
+    function ownerOfOverlay(bytes32 overlay) external view returns (address);
+
+    function stakeOfOverlay(bytes32 overlay) external view returns (uint256);
+}
 
 /**
  * @title Redistribution contract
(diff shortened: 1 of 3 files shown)


src/Redistribution.sol
```diff
@@ -398,19 +418,17 @@ contract Redistribution is AccessControl, Pausable {
...
                         overlay: currentCommits[i].overlay,
                         stake: currentCommits[i].stake,
-                        stakeDensity: currentCommits[i].stake * uint256(2**_depth),
+                        stakeDensity: currentCommits[i].stake * uint256(2 ** _depth),
                         hash: _hash,
                         depth: _depth
                     })
                 );
 
-                nonceBasedRandomness(_revealNonce);
```
significance, line 407: should we PR against the 0.4 tag?

src/Redistribution.sol
```diff
@@ -428,54 +446,40 @@ contract Redistribution is AccessControl, Pausable {
...
         require(cr > currentClaimRound, "round already received successful claim");
 
-        string memory truthSelectionAnchor = currentTruthSelectionAnchor();
-
-        uint256 currentSum;
         uint256 currentWinnerSelectionSum;
         bytes32 winnerIs;
         bytes32 randomNumber;
-
+        uint256 randomNumberTrunc;
```
significance, line 456: should this be here?

src/Redistribution.sol
```diff
@@ -360,22 +381,21 @@ contract Redistribution is AccessControl, Pausable {
...
+        // Check status of commit phase
         require(cr == currentCommitRound, "round received no commits");
+
+        // Set state for new reveal round on first reveal
         if (cr != currentRevealRound) {
             currentRevealRoundAnchor = currentRoundAnchor();
             delete currentReveals;
             currentRevealRound = cr;
             updateRandomness();
+            emit CurrentRevealAnchor(cr, currentRevealRoundAnchor);
```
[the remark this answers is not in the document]
n00b21337: It shouldn't make difference as updateRandomness doesn't affect cr or currentRevealRoundAnchor value. 
Also I put it according to placing in that MD from friday, not a problem to move it above if you think there is some reason, I cant see it though :)  (moved it in new PR)

src/Redistribution.sol
n00b21337: ok, lets do that.

src/Redistribution.sol
n00b21337: It was all reworked in the meantime but as agreed lets use exact v0.4 so it will be there.

n00b21337:
Closing this one as we want to apply all of this on v0.4.0 so this PR is not valid.

PR #133, opened 2023-04-03
fix: New events added

n00b21337:


src/Redistribution.sol
```diff
@@ -132,9 +128,9 @@ contract Redistribution is AccessControl, Pausable {
     }
 
     /**
-     * @dev Emitted when the winner of a round is selected in the claim phase.
+     * @dev Emitted when the winner of a round is selected in the claim phase with current round number
      */
-    event WinnerSelected(Reveal winner);
+    event WinnerSelected(Reveal winner, uint256 roundNumber);
```
significance, line 133: on second thoughts i think we should remove the round numbers - they are trivial to calculate from blockheight and leaving the events untouched will reduce the burden on users of the contract events api

scripts/deploy_redis.ts
```diff
@@ -0,0 +1,32 @@
...
+      '0x3e475aEAB162E28fee46E69225af446D3c4f3Bd3',
+    ];
+    const redisFactory = await ethers.getContractFactory('Redistribution');
+    console.log('Deploying contract...');
+    const redis = await redisFactory.deploy(...args);
+    await redis.deployed();
+    console.log(`Deployed contract to: ${redis.address}`);
+    await redis.deployTransaction.wait(6);
+
+    if (network.name == 'testnet' || network.name == 'sepolia') {
```
mrekucci, line 20: This if statement is always `true`; check line 6.

src/Redistribution.sol
```diff
@@ -132,7 +128,7 @@ contract Redistribution is AccessControl, Pausable {
     }
 
     /**
-     * @dev Emitted when the winner of a round is selected in the claim phase.
+     * @dev Emitted when the winner of a round is selected in the claim phase with current round number
```
nugaon, line 131: I guess you don't need this change anymore.

hardhat.config.ts
```diff
@@ -103,29 +103,44 @@ const config: HardhatUserConfig = {
...
       chainId: 100,
     },
   },
   etherscan: {
     apiKey: {
-      mainnet: '<gnosis-api-key>',
-      testnet: '<goerli-api-key>',
+      mainnet: mainnetEtherscanKey ? mainnetEtherscanKey : '',
+      testnet: testnetEtherscanKey ? testnetEtherscanKey : '',
+      sepolia: testnetEtherscanKey ? testnetEtherscanKey : '',
```
nugaon, line 143:
```suggestion
      mainnet: mainnetEtherscanKey || '',
      testnet: testnetEtherscanKey || '',
      sepolia: testnetEtherscanKey || '',
```

src/Redistribution.sol
```diff
@@ -636,6 +648,8 @@ contract Redistribution is AccessControl, Pausable {
 
         PostageContract.withdraw(winner.owner);
 
+        emit PostageStampStatus(PostageContract.validChunkCount(), PostageContract.pot());
```
nugaon, line 651: It should be before the withdraw operation since the pot will be `0` always.

nugaon (changes requested):

scripts/deploy_redis.ts
```diff
@@ -0,0 +1,31 @@
+import 'hardhat-deploy-ethers';
+import { ethers, network } from 'hardhat';
+import verify from '../utils/verify';
+
+async function main() {
+  if (network.name == 'testnet' || network.name == 'sepolia') {
```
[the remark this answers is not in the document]
n00b21337: sorry, this was just to test out things quickly, changed this part already.

hardhat.config.ts
n00b21337: nice

src/Redistribution.sol
n00b21337: good point, this wasn't really needed. just added it as it could be useful. will remove it as we wanted to get validchunk after withdrawl and clearing of stamps

zelig (approved):

significance (approved):

mrekucci (approved):

PR #134, opened 2023-04-04
feat: env oracle

nugaon:
This PR introduces the `EnvOracle` contract that has only one field, namely `minimumBeeVersion`.

`minimumBeeVersion` holds the minimum Bee client version required to play in the Redistribution game.
This value must be checked before participating in a round and forbid to send any transaction to the Redistribution contract if the used Bee version is lower than `minimumBeeVersion`.

This practice ensures that the Bee client is aligned with the breaking changes of the Redistribution contract.


n00b21337:
test comment for PR

n00b21337:
Have a few thoughts on this one. 
The code looks nice, concerns are conceptual, code is basically most about checking proper semver, which could be offloaded to offchain check in deployment code. Usually, that is good practice as onchain computing cost money.

Also I am wondering what would be the integration, do we want to integrate this code into redisitribution.sol or is this going to be just checked from bee node and if it doesn't pass the check they dont call other redis methods, which should be also fine. Maybe to simplify APIs and export for bee node we could use Interface to this contract and just add a call to version in redistribution contract so its just another method there. Would be simpler for them then mapping out new contract. Also if we would just check offchain semver it could just be actually one variable extra in the redistribution contract which would really simplify it all.

Also if there is some idea that we do more dependent integration on SC level, and each call of redis phase is checked on version, then this whole contract should be written as an abstract, and check should be a modifier.

So we could go with any of these choices IMO, its mostly dependent on what concept we want to push out and why. 
Wanted to raise this discussion before we finalize it.

nugaon:
> code is basically most about checking proper semver, which could be offloaded to offchain check in deployment code. Usually, that is good practice as onchain computing cost money.

IMO it is safer to handle this on-chain, since the chain can be seen as the backend-side of this architecture. 
On web2, admin methods also can be implemented on client side in many cases but that is still not a great practice. 
If Bee clients integrated EnvOracle and reacted on its events then a little mistake can cause failures that has monetary consequences as well.
I hope Swarm can cover the computational cost to execute this function couple of times : )     

> Also I am wondering what would be the integration, do we want to integrate this code into redisitribution.sol or is this going to be just checked from bee node and if it doesn't pass the check they dont call other redis methods, which should be also fine.

Latter, because this contract address should be fixed and the Redistribution and other contract addresses can vary.

n00b21337:
> > code is basically most about checking proper semver, which could be offloaded to offchain check in deployment code. Usually, that is good practice as onchain computing cost money.
> 
> IMO it is safer to handle this on-chain, since the chain can be seen as the backend-side of this architecture. On web2, admin methods also can be implemented on client side in many cases but that is still not a great practice. If Bee clients integrated EnvOracle and reacted on its events then a little mistake can cause failures that has monetary consequences as well. I hope Swarm can cover the computational cost to execute this function couple of times : )
>

yes, of course, actual  cost is not an issue here, it's more me as a solidity dev always thinking about making things less bloated and less expensive by default :)

I am also thinking about the process, actually when this changes, it will be changed through that multisig wallet, so I guess it makes sense that we don't make a mistake there in format, also would be really surprised we do as its rather simple variable to enter and check, so we would be really bad in entering and checking data if we do that :)
But yes, it's not an expensive thing to add and do and it would cost a lof more if we somehow make mistake and add wrong format, so ok lets do and be extra safe in that regard.
 
> > Also I am wondering what would be the integration, do we want to integrate this code into redisitribution.sol or is this going to be just checked from bee node and if it doesn't pass the check they dont call other redis methods, which should be also fine.
> 
> Latter, because this contract address should be fixed and the Redistribution and other contract addresses can vary.

This part is kind of obsolete, as when the redistribution addresses changes then the old contract is paused anyway so they cant use it.
The other thing is that we are aiming to proxy everything in some time so then redis addy would be the same.

Either way think we should add it at as interface to Redis as it would take less integration code for bee node.

n00b21337:
I will make a commit with suggestions on how we could make lighter integration to the current workflow.

n00b21337:
@nugaon check here a suggestion for integration.
https://github.com/ethersphere/storage-incentives/pull/134/commits/4614b819fa3405ab37e7c0ce81d64649b872ffaa

The reason for this will be that we don't need to change .sh script in /codegen  and also we don't need to add extra data to 007 step in the deploy process and adjust those 2 .json files in the root of the repo. This should also be easier for the bee node to just then call one method in ABI on the same contract and not need to map a new contract.

WDYT?

nugaon:
IMO, either we should handle this value in the Redistribution contract or keep it as a separated and not referenced smart contract.

Actually, that also can be a mechanism for bumping bee version to recognize the old Redis contract is not functioning anymore.

n00b21337:
> IMO, either we should handle this value in the Redistribution contract or keep it as a separated and not referenced smart contract.
> 
> Actually, that also can be a mechanism for bumping bee version to recognize the old Redis contract is not functioning anymore.

I am definitely more on the separation side, as redistribution contact is already big and bloated we don't need another thing added in there, also we could hit 24kb limit at some point and then we would need to again do some separation and refactoring. So I am all for this to be separated. Would like to hear arguments against integration through an interface as I don't see some at this point, but I would be fine with it if this is integrated with bee nodes directly. Not a problem IMO, just then this task needs more coding as it won't work without new go lang bindings, deployments exports, and what else might be missing for it. I made some doc here to list integrations demands, so please check here https://hackmd.io/ofFFKDh7S0aFV1f4n_yhvw

n00b21337:
To continue on this issue, we just have a method for checking but not enforcing these checks in the game.
I know its not bullet proof as ppl can fork and change the version but we should think about also implementing that on SC level, with bee nodes submitting their version on commits/reveals

More info on freezings here https://hackmd.io/hyIicg1YQAqMxbtVj_yyqg

n00b21337:
Per discussion on Matermost, this was decided that we won't use it in the end. The idea is that we just have a new redistribution contract that will force nodes to update and we won't have old nodes playing.

We can always rethink this down the line.

PR #139, opened 2023-04-05
add setters

n00b21337:


significance (approved):
lgtm aside from role mgmt - we should have a specific role for this, but fine for now and we can take care of this issue in a separate pr as there is more than one instance of this in current master

nugaon (approved):

PR #140, opened 2023-04-07
fix:redis deployment add missing part of script

n00b21337:
I didn't add proper scripting to create files for the bee node, so I am adding it to this PR. 
So in this script, we just change .json part for the redistribution contract which we update. It has already been done for testnet. So we can do RC with this PR one approved.


hardhat.config.ts
```diff
@@ -121,7 +121,7 @@ const config: HardhatUserConfig = {
       chainId: 31337,
     },
     testnet: {
-      url: PRIVATE_RPC_TESTNET ? PRIVATE_RPC_TESTNET : 'https://goerli.blockpi.network/v1/rpc/public',
+      url: PRIVATE_RPC_TESTNET ? PRIVATE_RPC_TESTNET : 'https://goerli.infura.io/v3/9aa3d95b3bc440fa88ea12eaa4456161',
```
mrekucci, line 124: Why is the account changed?

mrekucci (approved):

hardhat.config.ts
n00b21337: I find this RPC working a bit better

Issue #116, opened 2023-02-27
skipped rounds to be recast as 0 redundancy


significance:
at present, the price oracle does not amend the price in the event of skipped rounds (i.e. redundancy = 0)

in the event there are no applicants for a round, the round is 'skipped' and the the price oracle should increase the price by the inverse amount of redundancy = 8.

current price modulating numerator array:

`[0, 1036, 1027, 1025, 1024, 1023, 1021, 1017, 1012]`

proposed:

`[1036, 1031, 1027, 1025, 1024, 1023, 1021, 1017, 1012]`

as there are no transaction in the event there are no applicants, the price oracle should remember the last round number during which the price was modulated, and, when the next transaction happens, apply the relevant price change for each of the interstitial skipped rounds.

n00b21337:
Sounds fine to me. On that note, I might found a bug in the current Redistribution contract ties to Oracle and Stamp. 
Line `OracleContract.adjustPrice(uint256(k));` could be the problem
as in adjustPrice there is a
`require(redundancy > 0, "unexpected zero");`

which means if k is 0, then adjustPrice is reverted which also reverts the **claim** call in redistribution contract. I wonder if this is meant to be like that?

A similar pattern is with PostageContract.withdraw(winner.owner);  which has
require(ERC20(bzzToken).transfer(beneficiary, totalAmount), "failed transfer");
which also could fail, meaning again the whole **claim** process is failing and all the house keeping in it.

This can be fixed with try/catch pattern. Just need to know what do we exactly want and do we want to fail the claim in those cases or not?

significance:
in the future, the plan is to note skipped rounds and then apply a price increase for each one skipped

n00b21337:
Ok, that makes sense. About the withdraw, possibly it could fail but very unlikely. So we can keep it like it is now.

n00b21337:
My understanding of the setup is that we have this code, which prevents us from get into to setPrice when there are no reveals.
And if there are reveals there are commits and if there are commits then there is redundancy as its basically value of K which is the value of the number of LEGIT commits. So unless we have commits and reveals that are wrong we don't have a chance to set a price. Which seems to me that it could be a problem as we want to increase a price if there is no one commitng and claimaing as well as when there are wrong commits/claims,  thoughs @significance  ?

```
    function claim() external whenNotPaused {
        require(currentPhaseClaim(), "not in claim phase");
        uint256 cr = currentRound();
[...]

PR #157, opened 2023-06-15
feat: Skipped rounds recast

n00b21337:
PR for issue that is linked below

https://github.com/ethersphere/storage-incentives/issues/116


src/PriceOracle.sol
```diff
@@ -101,7 +108,10 @@ contract PriceOracle is AccessControl {
 
             uint256 multiplier = minimumPrice;
             uint256 usedRedundancy = redundancy;
+            uint256 currentRoundNumber = currentRound();
 
+            // price can only be adjusted once per round
+            require(currentRoundNumber != lastClaimedRound, "price already adjusted in this round");
```
ldeffenb, line 114: Should this be != or >?  There is logic below that will get a negative rounds skipped if the round is ever allowed to go backwards.

src/PriceOracle.sol
```diff
@@ -111,22 +121,42 @@ contract PriceOracle is AccessControl {
...
-
             // the multiplier is used to ensure whole number
+
+            // We first apply the increase/decrease rate for the current round
+            uint256 ir = increaseRate[usedRedundancy];
             currentPrice = (ir * currentPrice) / multiplier;
 
-            //enforce minimum price
+            // If previous rounds were skipped, use MAX price increase for the previouse rounds
+            if (skippedRounds > 0) {
```
ldeffenb, line 138: This would seem to protect against round number going backwards, but I think it would be best to reject it with a > check above.

src/PriceOracle.sol
```diff
@@ -111,22 +121,42 @@ contract PriceOracle is AccessControl {
...
+                }
+            }
+
+            // Enforce minimum price
             if (currentPrice < minimumPrice) {
                 currentPrice = minimumPrice;
             }
 
             postageStamp.setPrice(currentPrice);
+            lastClaimedRound = currentRoundNumber;
```
ldeffenb, line 151: For clarity, is it truly "lastClaimedRound" or in fact something like "lastAdjustedRound"?   We really don't know who/why adjustPrice was invoked.

ldeffenb (changes requested):

src/PriceOracle.sol
n00b21337: Primary purpose of this was to not let the possibility to call adjustPrice in the same round more than once, hence the message.

">" would do the same but if we manage to get "currentRoundNumber" to be lower than "lastClaimedRound"
we do have a set of some other problems :)

src/PriceOracle.sol
n00b21337: We could use > as a double insurance, yes.

src/Redistribution.sol
```diff
@@ -538,14 +538,18 @@ contract Redistribution is AccessControl, Pausable {
     function currentRoundAnchor() public view returns (bytes32 returnVal) {
         uint256 cr = currentRound();
 
+        // This will be called in reveal phase and set as currentRevealRoundAnchor or in
+        // commit phase when checking eligibility for next round by isParticipatingInUpcomingRound
         if (currentPhaseCommit() || (cr > currentRevealRound && !currentPhaseClaim())) {
             return currentSeed();
         }
 
+        // This will never be called  ¯\_(ツ)_/¯
```
nugaon, line 547: this is for sanity check

test/PriceOracle.test.ts
```diff
@@ -16,7 +16,8 @@ before(async function () {
   others = await getUnnamedAccounts();
 });
 
-const increaseRate = [0, 1036, 1027, 1025, 1024, 1023, 1021, 1017, 1012];
+const increaseRate = [1036, 1031, 1027, 1025, 1024, 1023, 1021, 1017, 1012];
+const roundLength = 152;
```
nugaon, line 20: this is already defined in the Redistribution test. would be nice to define it in utils instead (in one place).

test/Redistribution.test.ts
```diff
@@ -1,16 +1,17 @@
...
 const { read, execute } = deployments;
 const phaseLength = 38;
 const roundLength = 152;
 
-const increaseRate = [0, 1036, 1027, 1025, 1024, 1023, 1021, 1017, 1012];
+const increaseRate = [1036, 1031, 1027, 1025, 1024, 1023, 1021, 1017, 1012];
 
 const round2Anchor = '0xa6eef7e35abe7026729641147f7915573c7e97b47efa546f5f6e3230263bcb49';
 const round3AnchoIfNoReveals = '0xac33ff75c19e70fe83507db0d683fd3465c996598dc972688b7ace676c89077b';
+const round5Anchor = '0x17ef568e3e12ab5b9c7254a8d58478811de00f9e6eb34345acd53bf8fd09d3ec';
```
nugaon, line 14: not used variable

test/Redistribution.test.ts
```diff
@@ -830,6 +853,62 @@ describe('Redistribution', function () {
...
+          await mineNBlocks(phaseLength);
+
+          await r_node_5.reveal(overlay_5, depth_5, hash_5, reveal_nonce_5);
+          await r_node_6.reveal(overlay_6, depth_6, hash_6, reveal_nonce_6);
+          await mineNBlocks(phaseLength);
+
+          expect(await r_node_5.isWinner(overlay_5)).to.be.true;
+          expect(await r_node_6.isWinner(overlay_6)).to.be.false;
+
+          const tx2 = await r_node_6.claim();
```
nugaon, line 901: not used `tx2` constant

test/Redistribution.test.ts
```diff
@@ -830,6 +853,62 @@ describe('Redistribution', function () {
...
+        });
+
+        it('if both reveal, but after 4 skipped rounds, check proper price increase', async function () {
+          const nodesInNeighbourhood = 2;
+
+          // Check if the increase is properly applied, we have four skipped rounds here
+          const newPrice = (increaseRate[nodesInNeighbourhood] * price1) / 1024;
```
nugaon, line 908: a constant could be used instead of 1024

test/Redistribution.test.ts
```diff
@@ -62,6 +63,22 @@ const overlay_4 = '0xaedb2a8007316805b4d64b249ea39c5a1c4a9ce51dc8432724241f41ecb
...
+const depth_5 = '0x06';
+const reveal_nonce_5 = '0xb5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33';
+
+let node_6: string;
+const overlay_6 = '0x152d169abc6e6a0e0a2a7b78dcfea0bebe32942f05e9bb10ee2996203d5361ef';
+const stakeAmount_6 = '100000000000000000';
+const nonce_6 = '0xb5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33';
+const hash_6 = '0xb5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33';
+const depth_6 = '0x06';
+const reveal_nonce_6 = '0xb5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33';
```
nugaon, line 80: why was it necessary to add 2 more node definitions? why does not your test work with the already existing nodes?

nugaon (commented):
you can use lower depth or different nonce for committing instead of introducing new wallets.

test/PriceOracle.test.ts
n00b21337: yes, that would make sense. just with new values and in other PR

test/Redistribution.test.ts
n00b21337: That was done in the other PR for oracle.

test/Redistribution.test.ts
n00b21337: Because I wanted to test for a particular number of rounds that are skipped, and with those rounds current nodes overlay won't work.  Anchor will be out of depth for them as you probably know.
So I took the rounds I wanted to check for and found addresses and overlays that would work. 
It could be done differently with the same nodes but then I would need to mine a lot of rounds to get to those matching anchors, which would also slow down tests for this case but wouldn't test it any better than the current setup.

src/Redistribution.sol
nugaon: you can remove it, but then leave a comment with the clarification that the check tries to achieve.

test/Redistribution.test.ts
nugaon: that's why I let you know the two other approaches in the main comment of the review.

n00b21337: That just seemed like the much harder way to achieve the thing I wanted :)

I could use depth 0x00 but that seems like cheating the test and even 0x01 is not into the anchor I wanted to use. 
I don't want to change the anchor and mine 1000 blocks to find some matching anchor to current nodes as I am also trying to replicate some real-world scenario where a few rounds would be skipped, not thousands so I took that as constraints.

Let's say I want to use the same depth of 0x06 as the rest of the tests. I have an anchor that starts with "0xa92b32" 
How would you adjust the nonce of current nodes to match that anchor?

nugaon: it is not harder way IMO but cleaner for sure. we shouldn't add nodes for all new cases.

in what particular point your test cases depend on the depth to call 0 depth a cheat?

I use https://github.com/rndlabs/swarm-tools-rs/pull/7 for creating nonces

n00b21337: Because if we use depth 0x00 then we are not properly testing proximity and anchors/overlays which is part of the test.

n00b21337: will close this for now, If I make the time I will revisit this as it's not actually a problem, and it's more of tooling improvement which is of low importance.

src/PriceOracle.sol
```diff
@@ -24,24 +24,34 @@ contract PriceOracle is AccessControl {
...
+    uint256[] public increaseRate = [514191, 514182, 514173, 514164, 514155, 514146, 514137, 514128, 514119];
 
     uint16 targetRedundancy = 4;
     uint16 maxConsideredExtraRedundancy = 4;
 
     // When the contract is paused, price changes are not effective
     bool public isPaused = true;
 
+    // The length of a round in blocks.
+    uint256 public roundLength = 152;
```
mrekucci, line 43: More detailed description would be appreciated, e.g.: why it is needed, why it is equal to 152, how would increase/decrease affect the code, etc...

test/PriceOracle.test.ts
```diff
@@ -82,6 +83,9 @@ describe('PriceOracle', function () {
         await priceOracle.setPrice(minPriceString);
         price0SetBlock = await getBlockNumber();
 
+        // Set price base
+        const priceBase = await priceOracle.priceBase();
```
mrekucci, line 87: This issue reported by the checker should be addressed.

mrekucci (approved):

nugaon (approved):
`mineOverlaysInDepth` is not used (and seems wrong because does not return value), in `test: phase4` PR, the overlay mining is more optimal since it mines nonce instead of private key. I would remove this function. 

if you detail parameters next time, please use JSDoc format for the comments

test/PriceOracle.test.ts
n00b21337: fixed

n00b21337:
> `mineOverlaysInDepth` is not used (and seems wrong because does not return value), in `test: phase4` PR, the overlay mining is more optimal since it mines nonce instead of private key. I would remove this function.
> 
> if you detail parameters next time, please use JSDoc format for the comments

I added JSDoc now.

Please keep it for now as I used it and it did help me with some basic mining. When I try that other tool and see that it has all I need will remove it.

zelig (approved):

PR #159, opened 2023-06-20
style: stamp

n00b21337:
This is the basic reformat I want to do on all contracts. Many have just thrown around functions with no order.
There are 2 basic types functions, the one the READ state and the one that CHANGE the state.
Usually called setters and getters.  So good practice is to group functions in that manner, it's easier for review and flow to know where is what, similar is with type and variable ordering and events. I added namings where should be what section. No code logic has been changed here.

Best viewed as whole file  to get proper impression, here
https://github.com/ethersphere/storage-incentives/blob/d1e303a05b3163928c4bcf3f1663553a50606ded/src/PostageStamp.sol


src/PostageStamp.sol
```diff
@@ -159,6 +174,7 @@ contract PostageStamp is AccessControl, Pausable {
         // since the block the contract was deployed, so we must supplement this batch's
         // _initialBalancePerChunk with the currentTotalOutPayment()
         uint256 normalisedBalance = currentTotalOutPayment() + (_initialBalancePerChunk);
+        require(normalisedBalance > 0, "normalisedBalance cannot be zero");
```
n00b21337, line 177: This has been moved above so require is done sooner, saves a bit of gas if fails, and is better grouped with the actual value

src/PostageStamp.sol
```diff
@@ -110,6 +84,43 @@ contract PostageStamp is AccessControl, Pausable {
...
+     * @dev Emitted when the depth of an existing batch increases.
+     */
+    event BatchDepthIncrease(bytes32 indexed batchId, uint8 newDepth, uint256 normalisedBalance);
+
+    /**
+     *@dev Emitted on every price update.
+     */
+    event PriceUpdate(uint256 price);
+
+    // ----------------------------- CONSTRUCTOR ------------------------------
```
nugaon, line 122: I would remove this, it does not add much to anything, on the other hand introduces to obligation to write this above every constructor.

nugaon (approved):
LGTM

nugaon:
changed pr name since `reformat` is not a [conventional commit keyword](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#type-enum)

n00b21337:
> changed pr name since `reformat` is not a [conventional commit keyword](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#type-enum)

makes sense. lets use those convetions.

src/PostageStamp.sol
n00b21337: ahh, well its also a convention to use it. good solidity code will always have it. not really much overhead as you have one constructor per contract, so not much extra work, but always nice to have when scrolling as it is expected to be there if all all sections are commented.

PR #160, opened 2023-06-20
style: oracle

n00b21337:

view here
https://github.com/ethersphere/storage-incentives/blob/6c1002e5bca3caf8e40dbb5d9b4a91d9e82e5fe4/src/PriceOracle.sol

### DIFF
--- a/src/PriceOracle.sol
+++ b/src/PriceOracle.sol
@@ -13,10 +13,7 @@ interface IPostageStamp {
  */
 
 contract PriceOracle is AccessControl {
-    /**
-     *@dev Emitted on every price update.
-     */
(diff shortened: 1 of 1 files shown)


src/PriceOracle.sol
```diff
@@ -62,24 +72,6 @@ contract PriceOracle is AccessControl {
         emit PriceUpdate(currentPrice);
     }
 
-    /**
-     * @notice Pause the contract.
-     * @dev Can only be called by the admin role.
-     */
-    function pause() external {
```
nugaon, line 69: why was it necessary to move this function below? it didn't change the section.

PR #161, opened 2023-06-20
style: staking

n00b21337:

view here
https://github.com/ethersphere/storage-incentives/blob/61fb465c5bb3a1c6da88f0e4da71748e9c456352/src/Staking.sol


nugaon (approved):
LGTM!

PR #162, opened 2023-06-20
fix(Redis):  refactor and gas golfing

n00b21337:

view here
https://github.com/ethersphere/storage-incentives/blob/c5bfab1d29b06831e0a02ad9276c34cd4ef73927/src/Redistribution.sol


nugaon (approved):
would be nice to introduce corresponding prettier/solhint rules.

n00b21337:
> would be nice to introduce corresponding prettier/solhint rules.

Not sure what is asked here, but will try to elaborate on that.
There is a prettierrc.js file with solidity rules in repos root
which uses standard solidity formatting, also I would recommend setting it as default for formating. 
If using VS code, then best is to have it used "on Save" and as default formatter, like on image below

image

nugaon:
prettier/eslint should give at least warnings but even errors when the format of the document is not right.
it usually appears in the `problems` section in VS.
we could add an additional NPM style check command as well for this and run it in the test CI script, thereby it wouldn't be possible to merge if the code is not in the desired format.

nugaon:
created issue about it in #169

n00b21337:
> prettier/eslint should give at least warnings but even errors when the format of the document is not right. it usually appears in the `problems` section in VS. we could add an additional NPM style check command as well for this and run it in the test CI script, thereby it wouldn't be possible to merge if the code is not in the desired format.

yes, that is good idea.

n00b21337:
Merged #163  and #164  into this one as they were checked and approved.

n00b21337:
> > prettier/eslint should give at least warnings but even errors when the format of the document is not right. it usually appears in the `problems` section in VS. we could add an additional NPM style check command as well for this and run it in the test CI script, thereby it wouldn't be possible to merge if the code is not in the desired format.
> 
> yes, that is good idea.

Also, this is solved now.

n00b21337:
closed and continued here
https://github.com/ethersphere/storage-incentives/pull/186

PR #163, opened 2023-06-27
refactor:  Reveal flow

n00b21337:
Make the flow of reveal nicer to read and follow, finding the commit in the array does not need to be a loop for all the code below, it can be just set to ID and then use in the flow of the reveal code.

### DIFF
--- a/src/Redistribution.sol
+++ b/src/Redistribution.sol
@@ -5,7 +5,6 @@ import "@openzeppelin/contracts/security/Pausable.sol";
 import "./Util/TransformedChunkProof.sol";
 import "./Util/ChunkProof.sol";
 import "./Util/Signatures.sol";
-import "hardhat/console.sol";
 
 /**
  * Implement interfaces to PostageStamp contract, PriceOracle contract and Staking contract.
@@ -686,7 +685,6 @@ contract Redistribution is AccessControl, Pausable {
                 id = i;
             }
         }
-        console.log(id);
         return id;
     }
 
--- a/src/Redistribution.sol
+++ b/src/Redistribution.sol
@@ -313,7 +313,6 @@ contract Redistribution is AccessControl, Pausable {
      * @param _hash The reserve commitment hash.
      * @param _revealNonce The nonce used to generate the commit that is being revealed.
      */
-
     function reveal(bytes32 _overlay, uint8 _depth, bytes32 _hash, bytes32 _revealNonce) external whenNotPaused {
         require(currentPhaseReveal(), "not in reveal phase");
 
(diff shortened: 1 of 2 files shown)


src/Redistribution.sol
```diff
@@ -313,6 +313,7 @@ contract Redistribution is AccessControl, Pausable {
      * @param _hash The reserve commitment hash.
      * @param _revealNonce The nonce used to generate the commit that is being revealed.
      */
+
```
nugaon, line 316: delete this breakline

src/Redistribution.sol
```diff
@@ -679,6 +674,20 @@ contract Redistribution is AccessControl, Pausable {
...
 
+    /**
+     * @notice Helper function to get this node reveal in commits
+     * @dev
+     */
+    function findCommit(bytes32 _overlay, bytes32 _commitHash, uint256 _length) internal view returns (uint256 id) {
+        id = type(uint256).max;
+        for (uint256 i = 0; i < _length; i++) {
+            if (currentCommits[i].overlay == _overlay && _commitHash == currentCommits[i].obfuscatedHash) {
+                id = i;
```
nugaon, line 685:
```suggestion
                return id;
```
it does not worth to run the iteration after this point since the `overlay` should appear once in the array.

src/Redistribution.sol
```diff
@@ -328,45 +329,39 @@ contract Redistribution is AccessControl, Pausable {
...
-                    currentCommits[i].stake * uint256(2 ** _depth),
-                    _hash,
-                    _depth
-                );
-
-                return;
-            }
-        }
+        // Check that the commit exists,
+        require(id != type(uint256).max, "no matching commit or hash");
```
nugaon, line 335: IMO it would be nicer to throw an exception in the `findCommit` function instead, because this check requires to know that the findCommit function will set `id` to `type(uint256).max` if there is no match.

src/Redistribution.sol
```diff
@@ -679,6 +674,20 @@ contract Redistribution is AccessControl, Pausable {
...
+     * @dev
+     */
+    function findCommit(bytes32 _overlay, bytes32 _commitHash, uint256 _length) internal view returns (uint256 id) {
+        id = type(uint256).max;
+        for (uint256 i = 0; i < _length; i++) {
+            if (currentCommits[i].overlay == _overlay && _commitHash == currentCommits[i].obfuscatedHash) {
+                id = i;
+            }
+        }
+        return id;
```
nugaon, line 688: throw the exception here.

n00b21337: yes, all of these good ideas, much better code, and less gas consumption as well. thnx

PR #164, opened 2023-07-03
refactor: cleanup state variables in redis

n00b21337:
Cleaning up state variables in redistribution contract that are obsolete.


src/Redistribution.sol
```diff
@@ -380,12 +370,15 @@ contract Redistribution is AccessControl, Pausable {
     ) external whenNotPaused {
         winner = winnerSelection();
 
+        uint256 indexInRC1;
+        uint256 indexInRC2;
```
nugaon, line 374: why don't you just initialize these with the code below?

nugaon (approved):

PR #168, opened 2023-07-25
fix(oracle): change rate of price modulations 

n00b21337:
Was done on top of the branch where we have skipped rounds included.
Reworked tests to reflect the changes.

### DIFF
--- a/test/Redistribution.test.ts
+++ b/test/Redistribution.test.ts
@@ -11,7 +11,6 @@ const increaseRate = [514191, 514182, 514173, 514164, 514155, 514146, 514137, 51
 
 const round2Anchor = '0xa6eef7e35abe7026729641147f7915573c7e97b47efa546f5f6e3230263bcb49';
 const round3AnchoIfNoReveals = '0xac33ff75c19e70fe83507db0d683fd3465c996598dc972688b7ace676c89077b';
-const round5Anchor = '0x17ef568e3e12ab5b9c7254a8d58478811de00f9e6eb34345acd53bf8fd09d3ec';
 
 const maxInt256 = 0xffff; //js can't handle the full maxInt256 value
 
@@ -915,7 +914,7 @@ describe('Redistribution', function () {
           expect(await r_node_5.isWinner(overlay_5)).to.be.true;
           expect(await r_node_6.isWinner(overlay_6)).to.be.false;
 
-          const tx2 = await r_node_6.claim();
+          await r_node_6.claim();
         });
 
         it('if both reveal, but after 4 skipped rounds, check proper price increase', async function () {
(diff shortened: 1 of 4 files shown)


test/util/tools.ts
```diff
@@ -22,14 +22,18 @@ async function getBlockNumber(): Promise<number> {
   return parseInt(blockNumber);
 }
 
-async function skippedRoundsIncrease(n: number, price: number): Promise<number> {
-  const maxIncreaseRate = 1036;
-  const multiplier = 1024;
+async function skippedRoundsIncrease(
+  n: number,
+  price: number,
+  priceBaseString: number,
```
nugaon, line 28: why is it named `priceBaseString` when it is a number?

test/util/tools.ts
```diff
@@ -22,14 +22,18 @@ async function getBlockNumber(): Promise<number> {
...
+  maxIncreaseRate: number
+): Promise<number> {
+  const base = priceBaseString;
```
nugaon, line 31: just called it `base` in the argument definitions, this assignment does not do anything than just renaming the variable

test/Redistribution.test.ts
```diff
@@ -688,6 +689,10 @@ describe('Redistribution', function () {
           r_node_1 = await ethers.getContract('Redistribution', node_1);
           r_node_2 = await ethers.getContract('Redistribution', node_2);
 
+          // Set price base
+          const priceBase = await priceOracle.priceBase();
+          priceBaseString = priceBase.toString();
```
nugaon, line 694: it is string by default, isn't it?

test/Redistribution.test.ts
```diff
@@ -873,6 +883,10 @@ describe('Redistribution', function () {
           priceOracle = await ethers.getContract('PriceOracle', deployer);
           await priceOracle.unPause(); // TODO: remove when price oracle is not paused by default.
 
+          // Set price base
+          const priceBase = await priceOracle.priceBase();
+          priceBaseString = priceBase.toString();
```
nugaon, line 888: you cast it to int at each occurrence, why don't you just cast it here to int instead?

nugaon (approved):
Overall the main changes seem correct

test/Redistribution.test.ts
n00b21337: was modeled after
 let minPriceString: string;
 
 as it derived from there as a new separate value. That needed a string for setPrice method, but this one doesn't and you are right, it shouldn't be cast back and forth. I will improve that part. thnx

test/Redistribution.test.ts
n00b21337: It is not, it's an Object. BigNumber object, so it needs to be cast down to something but this was not needed to be string as we just use the number, so good call on this

n00b21337:
> Overall the main changes seem correct

thnx for the review, good call on tests, the code was sub-optimal and also had some bad namings. Improved all of that and more, tests run smoothly but are more clear. Will merge tomorrow if you do not have anything other to add.

mrekucci (approved):
Except for the reported lint issues about unused variables. LGTM!

PR #171, opened 2023-08-07
fix(redis): Gas golfing

n00b21337:
Gas golfing for refactored redistribution contract

### DIFF
--- a/src/Redistribution.sol
+++ b/src/Redistribution.sol
@@ -369,6 +369,7 @@ contract Redistribution is AccessControl, Pausable {
 
         uint256 indexInRC1;
         uint256 indexInRC2;
+        bytes32 currentAnchor = currentRevealRoundAnchor;
 
         // rand(14)
         indexInRC1 = uint256(seed) % 15;
@@ -378,26 +379,17 @@ contract Redistribution is AccessControl, Pausable {
             indexInRC2++;
         }
 
-        require(
-            inProximity(entryProofLast.proveSegment, currentRevealRoundAnchor, winner.depth),
-            "witness is not in depth"
-        );
+        require(inProximity(entryProofLast.proveSegment, currentAnchor, winner.depth), "witness is not in depth");
         inclusionFunction(entryProofLast, 30);
         stampFunction(entryProofLast);
         socFunction(entryProofLast);
 
-        require(
-            inProximity(entryProof1.proveSegment, currentRevealRoundAnchor, winner.depth),
-            "witness is not in depth"
-        );
+        require(inProximity(entryProof1.proveSegment, currentAnchor, winner.depth), "witness is not in depth");
         inclusionFunction(entryProof1, indexInRC1 * 2);
         stampFunction(entryProof1);
         socFunction(entryProofLast);
 
-        require(
-            inProximity(entryProof2.proveSegment, currentRevealRoundAnchor, winner.depth),
-            "witness is not in depth"
-        );
+        require(inProximity(entryProof2.proveSegment, currentAnchor, winner.depth), "witness is not in depth");
         inclusionFunction(entryProof2, indexInRC2 * 2);
         stampFunction(entryProof2);
         socFunction(entryProofLast);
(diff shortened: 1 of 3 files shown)


src/Redistribution.sol
```diff
@@ -141,41 +141,41 @@ contract Redistribution is AccessControl, Pausable {
     // Reveals for the current round.
     Reveal[] public currentReveals;
 
-    // Role allowed to pause.
-    bytes32 public constant PAUSER_ROLE = keccak256("PAUSER_ROLE");
-
     // Maximum value of the keccack256 hash.
-    bytes32 MaxH = bytes32(0x00000000000000000000000000000000ffffffffffffffffffffffffffffffff);
+    bytes32 private MaxH = bytes32(0x00000000000000000000000000000000ffffffffffffffffffffffffffffffff);
```
mfw78, line 145: Shouldn't this also be `constant` ?

n00b21337: yes absolutely, cleaned and added in latest commit.

n00b21337:
[gas_diff.txt](https://github.com/ethersphere/storage-incentives/files/12320891/gas_diff.txt)

Latest DIFF file for comparing optimization differences.

mrekucci:
There are some lint issues reported that should be fixed.

src/Redistribution.sol
```diff
@@ -622,14 +693,14 @@ contract Redistribution is AccessControl, Pausable {
      * @notice The number of the current round.
      */
     function currentRound() public view returns (uint256) {
```
nugaon, line 695: you changed all round related variables to `uint32` from `uint256`. why did left this out?

src/Redistribution.sol
```diff
@@ -360,39 +407,46 @@ contract Redistribution is AccessControl, Pausable {
...
         ChunkInclusionProof calldata entryProofLast
     ) external whenNotPaused {
-        winner = winnerSelection();
+        winnerSelection();
 
+        if (winner.owner != msg.sender) {
+            revert SenderNotWinner();
+        }
+
+        Reveal memory winnerSelected = winner;
```
nugaon, line 416: why not before the if statement?
```suggestion
       Reveal memory winnerSelected = winner;

        if (winnerSelected.owner != msg.sender) {
            revert SenderNotWinner();
        }
```

src/Redistribution.sol
```diff
@@ -594,10 +669,6 @@ contract Redistribution is AccessControl, Pausable {
             return currentSeed();
         }
 
-        if (currentPhaseReveal() && cr == currentRevealRound) {
-            require(false, "can't return value after first reveal");
```
nugaon, line 598: you just removed this but it was defined here because it has the same ambiguity as invoking `isParticipatingInUpcomingRound` in the reveal phase, hasn't it?

src/Redistribution.sol
```diff
@@ -931,17 +1037,27 @@ contract Redistribution is AccessControl, Pausable {
 
     function checkOrder(uint256 a, uint256 b, bytes32 trA1, bytes32 trA2, bytes32 trALast) internal pure {
         if (a < b) {
-            require(uint256(trA1) < uint256(trA2), "random element order check failed");
-            require(uint256(trA2) < uint256(trALast), "last element order check failed");
+            if (uint256(trA1) >= uint256(trA2)) {
+                revert RandomCheckFailed();
```
nugaon, line 1041: maybe `RandomCheckOrderFailed`?

nugaon (commented):
LGTM overall! : )

Merge it after the test: phase4 PR please.

src/Redistribution.sol
n00b21337: Generally, I changed the storage variables so we save gas, these are view and/or memory ones that don't change much for the optimization of gas if I change it to a lower unit.

But the main thing is that cr and currentRound() are used in many places so I would need to downcast a lot of other variables and how other functions work or upcast it again from 32 to 256 to keep them how they work now. Was afraid I could introduce some errors that way. 
It could be done, just need to be careful not to mess up something and I didn't get to that so far and take it as low priority.

src/Redistribution.sol
```diff
@@ -304,48 +341,58 @@ contract Redistribution is AccessControl, Pausable {
...
-            inProximity(currentCommits[id].overlay, currentRevealRoundAnchor, _depth),
-            "anchor out of self reported depth"
-        );
+
+        if (!inProximity(currentCommit.overlay, currentRevealRoundAnchor, _depth)) {
+            revert OutOfDepth();
+        }
         // Check that the commit has not already been revealed
-        require(currentCommits[id].revealed == false, "participant already revealed");
+        if (currentCommit.revealed == true) {
```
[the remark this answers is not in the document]
n00b21337: Fixed.

src/Redistribution.sol
n00b21337: It could be, generally, we want to **revert** to fire ASAP as we save a bit of gas for trx that way, so just a good way to put it there before all others. Its small difference in the end but I don't see reason to put it before?

src/Redistribution.sol
n00b21337: I thought about this way more than I should now :) Going back and forth and putting it back and then again removing it.

So currentRoundAnchor is used here and with this it will never fire that condition.
image

For isParticipatingInUpcomingRound  I am not sure why do we want to have it there.  What this does is that we can check anchor in reveal round but before first reveal. And with that isParticipatingInUpcomingRound can be checked in reveal round before first reveal. I am wondering why we want to enable checking anchor and isParticipatingInUpcomingRound at the point before first reveal in reveal phase?

Seems obsolete to me. But maybe there is some case for it in bee node somehow?
(putting it back again now just in case I dont break bee node, but still wondering why do we need this? )

n00b21337: Ok I am trying to get why this is needed in that situation. Could be I dont know some info, seems strange to me to check for participation in the upcoming round in the revealed round running.  I added it, but tried to make it more into context as it is used for isParticipatingInUpcomingRound for some still unknown reason to me, so I added it there as check.

n00b21337: Hmm starting to get what they wanted with this. To not show next anchor in reveal phase, as it is set for next round after first reveal with new seed. I think it kind of convoluted this way. Will see If i can make it more clear and still keep the same idea.

n00b21337: I made this and just realized, this also doesnt make any sense as it will never get to that point, there is FIRST IF that will revert and will not get to that. So there is really no point in that code.

image

src/Redistribution.sol
nugaon: it just wrong in this current way. RandomCheck can be everything.

src/Redistribution.sol
nugaon: because it will read out from storage which is more expensive, isn't it? the memory read is 3-5 gas, the storage read is 200-800

PR #181, opened 2023-09-01
fix: depth

n00b21337:


src/Redistribution.sol
```diff
@@ -697,19 +781,40 @@ contract Redistribution is AccessControl, Pausable {
...
     }
 
+    /**
+     * @notice Returns minimum depth reveal has to have to participate in this round
+     */
+    function currentMinimumDepth() public view returns (uint8) {
+        // We are checking this in  reveal phase, in this phase have new commit round but
+        // for claim round it should be still the last time pot was claimed, which essentialy gives
+        // us how many rounds have been skipped. Because of the difference by default we substract
+        // 1 from the last winner depth
```
nugaon, line 802: last time we agreed on keeping the same depth for the next subsequent round and only deducting it after if there was a missed round. in this current form, as you wrote, the depth requirement immediately looser with 1.

nugaon: also, try to rephrase the text because it doesn't have the correct grammar form. It is just my feelings, I also use deepl.com for writing text sometimes.

src/Redistribution.sol
```diff
@@ -697,19 +781,40 @@ contract Redistribution is AccessControl, Pausable {
...
+
+        uint32 skippedRounds = currentCommitRound - currentClaimRound;
+        uint8 lastWinnerDepth = winner.depth;
+
+        // Minimum depth returned is 1, otherwise return the difference
```
nugaon, line 807: why is it 1? it should be 0 since that is the minimum viable storage depth

src/Redistribution.sol
```diff
@@ -931,17 +1057,27 @@ contract Redistribution is AccessControl, Pausable {
...
+                revert LastCheckFailed();
+            }
         } else {
-            require(uint256(trA2) < uint256(trA1), "random element order check failed");
-            require(uint256(trA1) < uint256(trALast), "last element order check failed");
+            if (uint256(trA2) >= uint256(trA1)) {
+                revert RandomCheckFailed();
+            }
+            if (uint256(trA1) >= uint256(trALast)) {
+                revert LastCheckFailed();
```
nugaon, line 1071: https://github.com/ethersphere/storage-incentives/pull/171#discussion_r1307459231

nugaon (changes requested):
Please, next time when you do PRs, do not include other branch's changes.

IMO this could be a feat PR instead.

src/Redistribution.sol
```diff
@@ -115,6 +92,7 @@ contract Redistribution is AccessControl, Pausable {
         bytes signature;
         bytes32 chunkAddr;
         bytes32 postageId;
+        uint64 chunkSpan;
```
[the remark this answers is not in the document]
n00b21337: yes, this was mistake. will redo it now as this is really chatoic.

n00b21337:
> Please, next time when you do PRs, do not include other branch's changes.
> 
> IMO this could be a feat PR instead.

yes, this was mistake. will redo it now as this is really chatoic.

src/Redistribution.sol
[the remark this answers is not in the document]
n00b21337: I reordered struct because that was part of memory packing optimization.
I do not have any idea what you are referring to as breaking data, don't think this is the case, please provide more info here.

PR #182, opened 2023-09-05
fix: depth check

n00b21337:


src/Redistribution.sol
```diff
@@ -482,6 +486,21 @@ contract Redistribution is AccessControl, Pausable {
...
+    /**
+     * @notice Returns minimum depth reveal has to have to participate in this round
+     */
+    function currentMinimumDepth() public view returns (uint8) {
+        // We are checking value in reveal phase, as the currentCommitRound is set to the current round
+        // but the currentClaimRound is still set to the last time claim was made
+        // We add 1 to ensure that for the next round the minimum depth is the same as last winner depth
+
+        uint8 skippedRounds = uint8(currentCommitRound - currentClaimRound) + 1;
+        uint8 lastWinnerDepth = winner.depth;
```
nugaon, line 498: can it be used in the first run as well? because then no claim happened to set a winner. is the declaration of the `winner` initializes all properties with def values?

src/Redistribution.sol
```diff
@@ -482,6 +486,21 @@ contract Redistribution is AccessControl, Pausable {
...
 
+    /**
+     * @notice Returns minimum depth reveal has to have to participate in this round
+     */
+    function currentMinimumDepth() public view returns (uint8) {
+        // We are checking value in reveal phase, as the currentCommitRound is set to the current round
+        // but the currentClaimRound is still set to the last time claim was made
+        // We add 1 to ensure that for the next round the minimum depth is the same as last winner depth
+
+        uint8 skippedRounds = uint8(currentCommitRound - currentClaimRound) + 1;
```
nugaon, line 497: IMHO we can accept that after 256 missing rounds the last winner.depth is counting down again (roughly 54 days)

nugaon (commented):
For each error we have now a test maybe this also require one. 

I still think though this PR should be a feat.

Thanks for cleaning up (and fixing the issues).
Nevertheless, please turn off your auto-formatter and only commit style related changes when you do a style PR. it is just too much always to check everything remained the same + filter out the real changes.

src/Redistribution.sol
n00b21337: yes the first run is case that should be checked with testing for sure.
what should happen is that 
 winner.depth = 0
 currentClaimRound = 0
 currentCommitRound = 198902 or something so  it would be true on last condition and will be 0

src/Redistribution.sol
n00b21337: I don't understand what was said here. :) Can you please try to explain again?

nugaon: just a note because of the possible overflow, it does not require any action. after 256 rounds the substraction will result in 0 because of the uint8 casting

nugaon: but 256 missing rounds unlikely, hopefully never happen

n00b21337: It would wrap around, meaning something like if there is 300 in difference it would be 44 as result.
So I am suggesting this as solution to make some upperLimit that can be reached

```
        uint256 difference = currentCommitRound - currentClaimRound;
        uint8 skippedRounds = uint8(difference > 254 ? 254 : difference) + 1;
```

n00b21337:
Continued here 
https://github.com/ethersphere/storage-incentives/pull/187

PR #189, opened 2023-09-25
refactor: claim interface

nugaon:
Refactored the claim interface where the postage related data got separated `postageProof` attribute in the claim function and renamed `socProofAttached` to `socProof`.
One unnecessary attribute got removed for postage proof which is the `chunkAddr`. Its value can be accessed either from `proveSegment` or `socProof.chunkAddr` at CAC or SOC respectively.


n00b21337 (approved):
Nothing to add here.

PR #198, opened 2023-10-02
fix: events and errors info

n00b21337:


src/Redistribution.sol
```diff
@@ -224,6 +229,7 @@ contract Redistribution is AccessControl, Pausable {
     error NoMatchingCommit(); // No matching commit and hash
     error NotClaimPhase(); // Game is not in the claim phase
     error NoReveals(); // Round did not receive any reveals
+    error FirsRevealDone(); // We don't want to return value after first reveal
```
ldeffenb, line 232: Missing a "t" on First?

ldeffenb (changes requested):
Just a minor spelling correction suggestion.

src/Redistribution.sol
n00b21337: solved, thnx

PR #200, opened 2023-10-03
fix: Gas golf stamp

n00b21337:


n00b21337:
doing it again in new PR and applying code on top of current with  in it

        //update validChunkCount to remove currently expired batches
        expireLimited(type(uint256).max);

PR #218, opened 2023-10-18
fix(redis): stamp alive check

nugaon:
`stampFunction` had a stricter condition for the alive check of stamp so that mined chunks cannot be signed with short-living cheap postage batch. This condition is not that strict because a dishonest actor can keep up a single batch for this action which is not a significant expense.

### DIFF
--- a/src/PostageStamp.sol
+++ b/src/PostageStamp.sol
@@ -392,12 +392,12 @@ contract PostageStamp is AccessControl, Pausable {
         emit PriceUpdate(_price);
     }
 
-    function setMinimumValidityBlocks(uint256 _value) external {
+    function setMinimumValidityBlocks(uint64 _value) external {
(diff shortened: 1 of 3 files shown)


n00b21337 (approved):
looks good, but needs test adjustment to not fail

n00b21337 (commented):
hmm yeah, I thought about this again 
So these are my objections
First, it will not be less than 0 as its returns uint256,
also, its little chance we are going to hit  this == 0 at the exact moment we check.

What will actually happen is that
PostageContract.remainingBalance(entryProof.postageProof.postageId)
will either be a positive number or it will revert

So maybe this would be the more precise code to use

```
        if (batchOwner == address(0)) {
            revert BatchDoesNotExist(); // Batch does not exist or expired
        }
```

PR #223, opened 2023-10-20
feat(redis): set sample max value

nugaon:
in case of Bee reserve size would raise in the future, we may need to adjust the reserve estimation for the new size.


n00b21337 (approved):
good

PR #227, opened 2023-10-26
feat:oracle proper implementation

n00b21337:


src/PriceOracle.sol
```diff
@@ -117,29 +121,35 @@ contract PriceOracle is AccessControl {
             uint32 _currentPrice = currentPrice;
             uint32 _minimumPrice = minimumPrice;
             uint32 _priceBase = priceBase;
+            uint64 _currentPriceUpscaled = currentPriceUpscaled;
```
nugaon, line 124: should be uint256

nugaon (commented):
I imagined this like we calculate bitshifted values in the `fields` and in accessor we shift to the  real `currentPrice`. 
will it keep the subtle changes?

src/PriceOracle.sol
n00b21337: 24577265
24578108
24578529
24578529
some example values for it above

it is 8 digit number that is in
_currentPriceUpscaled

uint64 has 20 digits

p.s.
also if you worry about calculation maximums it goes around this number
1.2886892e+13  before division.  So there is still 6 digits left to spare and its a safe type to operate on the value.

src/PriceOracle.sol
```diff
@@ -25,16 +25,18 @@ contract PriceOracle is AccessControl {
...
+    uint32 public minimumPrice = 24000;
 
     // The priceBase to modulate the price
-    uint32 public priceBase = 514155;
+    uint32 public priceBase = 524288;
 
     // The current price is the atomic unit.
     uint32 public currentPrice = minimumPrice;
 
+    uint32 private currentPriceUpscaled = minimumPrice << 10; // we upscale it by 2^10
```
nugaon, line 36: make it uint64 as you use it in other places.

src/PriceOracle.sol
```diff
@@ -117,29 +120,35 @@ contract PriceOracle is AccessControl {
...
             if (skippedRounds > 0) {
-                ir = increaseRate[0];
+                _changeRate = changeRate[0];
                 for (uint64 i = 0; i < skippedRounds; i++) {
-                    _currentPrice = (ir * _currentPrice) / _priceBase;
+                    _currentPriceUpscaled = (_changeRate * _currentPriceUpscaled) / _priceBase;
                 }
             }
 
+            _currentPrice = uint32(_currentPriceUpscaled) >> 10;
```
nugaon, line 140: you cast it to `uint32` then shift it? it should be the other way around

nugaon (changes requested):
So when we discussed this, the idea was that giving a separate accessor for `currentPrice` which is already bitshifted by 10, and the accessor takes this value and shift it back by 10. instead, we have two variables now with which we need to juggle.

n00b21337:
> So when we discussed this, the idea was that giving a separate accessor for `currentPrice` which is already bitshifted by 10, and the accessor takes this value and shift it back by 10. instead, we have two variables now with which we need to juggle.

but then values between will be lost.
this way we keep values in decimal 
iike 24000.3587 in one round then in next
24000.6788 and then finally in last we have
24001.046
if i dont keep it will be just 24000

src/PriceOracle.sol
n00b21337: Actually, this is probably not needed uint32 has 10 digits and our currentPriceUpscaled is usually around 8. But we could just give it more safe room for future and less worry.

src/PriceOracle.sol
n00b21337: As in the comment above, _currentPriceUpscaled is max 8 digits so when price raises 100x then we could have a problem. So it doesnt metter in this case, nothing would be lost, but I do agree its kind of better to have first bitshift then downcast.

If you wonder why did I use uint64 on other place it is the multiplication that has this numbers actually
 uint64 res1 = (524297 * 24576000) / 524288;
 
and in this case at one point it goes over uint32, so not to get problems I used uint64.

src/PriceOracle.sol
```diff
@@ -25,16 +25,15 @@ contract PriceOracle is AccessControl {
     uint64 public lastAdjustedRound;
 
     // The minimum price allowed
-    uint32 public minimumPrice = 1024;
+    uint32 public minimumPrice = 24000;
```
nugaon, line 28: I would do the same as for `currentPrice`

test/PostageStamp.test.ts
```diff
@@ -88,13 +91,18 @@ describe('PostageStamp', function () {
     let postageStampStamper: Contract, token: Contract, priceOracle: Contract;
     let batch: Batch;
     let batchSize: number, transferAmount: number;
-    const price0 = 1024;
-    let setPrice0Block: number;
+    let minimumPrice: number;
+    let price0: number;
+    let initialBlock: number;
```
nugaon, line 96: why was the old name not good enough? it was really clear what was it about, now it is not that straightforward.

nugaon (approved):
so now we have currentPriceUpscaled as we discussed, but it could be applied to minimumPrice as well.

test/PostageStamp.test.ts
n00b21337: hmm ok, probably right. will put it back.

src/PriceOracle.sol
n00b21337: will do that, makes sense, and looks better. thnx

PR #245, opened 2024-04-10
feat: Staking optimization, code and gas golfing

n00b21337:
Change a structure of Staking and also add gas optimizations, custom errors etc


nugaon (approved):
outside of my comment below LGTM

PR #252, opened 2024-04-16
New staking - part 1 and 2 - SWIP-19 and 20 combined

n00b21337:
This implements the feature allowing staked nodes to change their overlay without restaking (called neighbourhood hopping, see [SWIP-19](https://github.com/ethersphere/SWIPs/pull/48))

Implemented by changing stake registry to be keyed with node address instead of node overlay.


src/Staking.sol
```diff
@@ -89,84 +93,100 @@ contract StakeRegistry is AccessControl, Pausable {
...
+        stakes[_owner] = Stake({
             owner: _owner,
             overlay: overlay,
             stakeAmount: updatedAmount,
             lastUpdatedBlockNumber: block.number,
             isValue: true
         });
 
-        emit StakeUpdated(overlay, updatedAmount, _owner, block.number);
+        emit StakeUpdated(_owner, updatedAmount, stakes[_owner].overlay, block.number);
```
nugaon, line 117:
```suggestion
        emit StakeUpdated(_owner, updatedAmount, overlay, block.number);
```

src/Staking.sol
```diff
@@ -89,84 +93,100 @@ contract StakeRegistry is AccessControl, Pausable {
...
+     */
+    function changeOverlay(address _owner, bytes32 _nonce) external {
+        if (_owner != msg.sender) revert Unauthorized();
+        bytes32 overlay = keccak256(abi.encodePacked(_owner, reverse(NetworkId), _nonce));
+
+        if (stakes[_owner].isValue) {
+            stakes[_owner].overlay = overlay;
+            stakes[_owner].lastUpdatedBlockNumber = block.number;
+        }
+        emit ChangedOverlay(_owner, overlay);
```
nugaon, line 189: it changes only within the `if` block above.

src/Staking.sol
```diff
@@ -196,45 +216,53 @@ contract StakeRegistry is AccessControl, Pausable {
...
     /**
-     * @dev Returns the eth address of the owner of `overlay`.
-     * @param overlay Overlay of node
+     * @dev Returns the currently used overlay of the address.
+     * @param _owner address of node
      */
-    function ownerOfOverlay(bytes32 overlay) public view returns (address) {
-        return stakes[overlay].owner;
+    function ownerOfAddress(address _owner) public view returns (address) {
+        return stakes[_owner].owner;
```
nugaon, line 265: is there any situation when this is not equal to `_owner`?

test/Staking.test.ts
```diff
@@ -43,6 +43,14 @@ const updatedStakeAmount = '10000000000633633';
 const twice_stakeAmount_0 = '20000000000000000';
 const nonce_0 = '0xb5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33';
 
+let staker_1: string;
+const overlay_1 = '0xa6f955c72d7053f96b91b5470491a0c732b0175af56dcfb7a604b82b16719406';
```
nugaon, line 47: not used variable

test/Staking.test.ts
```diff
@@ -43,6 +43,14 @@ const updatedStakeAmount = '10000000000633633';
...
+const overlay_1_n_25 = '0x676766bbae530fd0483e4734e800569c95929b707b9c50f8717dc99f9f91e915';
+const stakeAmount_1 = '100000000000000000';
+const nonce_1 = '0xb5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33';
+const nonce_1_n_25 = '0x00000000000000000000000000000000000000000000000000000000000325dd';
+const stakeAmount_1_n_25 = '200000000000000000';
```
nugaon, line 52: not used variable

test/Staking.test.ts
```diff
@@ -408,4 +417,32 @@ describe('Staking', function () {
...
+    let sr_staker_1: Contract;
+
+    beforeEach(async function () {
+      await deployments.fixture();
+      token = await ethers.getContract('TestToken', deployer);
+      stakeRegistry = await ethers.getContract('StakeRegistry');
+
+      sr_staker_1 = await ethers.getContract('StakeRegistry', staker_1);
+
+      await mintAndApprove(staker_1, stakeRegistry.address, stakeAmount_1);
```
nugaon, line 431:
```suggestion
      sr_staker_1 = await ethers.getContract('StakeRegistry', staker_1);

      await mintAndApprove(staker_1, sr_staker_1.address, stakeAmount_1);
```
`stakeRegistry` is not really required

src/Staking.sol
```diff
@@ -23,14 +22,14 @@ contract StakeRegistry is AccessControl, Pausable {
         uint256 stakeAmount;
         // Block height the stake was updated
         uint256 lastUpdatedBlockNumber;
-        // Owner of `overlay`
+        // Owner of stake, we need to store owner again to be retrievable
```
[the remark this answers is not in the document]
n00b21337: Hmm that is good question that I been asking myself. So  if we actually want to get this

```
        Stake memory stake = stakes[_owner];
        if (stake.owner != msg.sender) revert Unauthorized();
```

I need to have owner as I cant get the key itself from the mapping itself but this is just not logical as I know the owner from function param then I can just do this

```
    function withdrawFromStake(address _owner, uint256 _amount) external whenPaused {
        Stake memory stake = stakes[_owner];
        if (_owner != msg.sender) revert Unauthorized();
```

What I am afraid of if for some reason we somehow do need to get key value from mapping we cant unless we save it like this as there.  But logic itself is that even if I cant do something like sakes["key"]  I don't need to as whichever key I might want to check is the value the I am trying to check in the first place so I don't need to get it from the mapping.

thinking out loud on this, hope it made sense, please confirm if you agree on it find some logic flaw.

I will remove it as I think we actually don't need it after this rethinking.

src/Staking.sol
n00b21337: yes, makes sense

src/Staking.sol
```diff
@@ -89,84 +91,99 @@ contract StakeRegistry is AccessControl, Pausable {
      * @notice Create a new stake or update an existing one.
      * @dev At least `_initialBalancePerChunk*2^depth` number of tokens need to be preapproved for this contract.
      * @param _owner Eth address used for overlay calculation.
-     * @param nonce Nonce that was used for overlay calculation.
-     * @param amount Deposited amount of ERC20 tokens.
+     * @param _nonce Nonce that was used for overlay calculation.
+     * @param _amount Deposited amount of ERC20 tokens.
      */
-    function depositStake(address _owner, bytes32 nonce, uint256 amount) external whenNotPaused {
+    function depositStake(address _owner, bytes32 _nonce, uint256 _amount) external whenNotPaused {
```
zelig, line 97: if this allows new nonce (change overlay), why do we need the `changeOverlay` function?

zelig, line 97: should eliminate owner argument and authorisation check and just use `msg.sender` as owner

src/Staking.sol
```diff
@@ -89,84 +91,99 @@ contract StakeRegistry is AccessControl, Pausable {
...
         }
 
         if (!ERC20(bzzToken).transfer(msg.sender, withDrawLimit)) revert TransferFailed();
     }
 
     /**
      * @dev Freeze an existing stake, can only be called by the redistributor
-     * @param overlay the overlay selected
-     * @param time penalty length in blocknumbers
+     * @param _owner the addres selected
```
zelig, line 142: here it is necessary

src/Staking.sol
```diff
@@ -89,84 +91,99 @@ contract StakeRegistry is AccessControl, Pausable {
...
+        bytes32 overlay = keccak256(abi.encodePacked(_owner, reverse(NetworkId), _nonce));
+
+        if (stakes[_owner].isValue) {
+            stakes[_owner].overlay = overlay;
+            stakes[_owner].lastUpdatedBlockNumber = block.number;
+            emit ChangedOverlay(_owner, overlay);
+        }
     }
 
     function changeNetworkId(uint64 _NetworkId) external {
```
zelig, line 189: this should be part of the deposit function and should be freely changeable by the owner I think

src/Staking.sol
```diff
@@ -196,45 +213,45 @@ contract StakeRegistry is AccessControl, Pausable {
...
-     * @dev Checks to see if `overlay` is frozen.
-     * @param overlay Overlay of staked overlay
+     * @dev Checks to see if `address` is frozen.
+     * @param _owner owner of staked address
      *
      * Returns a boolean value indicating whether the operation succeeded.
      */
-    function overlayNotFrozen(bytes32 overlay) internal view returns (bool) {
-        return stakes[overlay].lastUpdatedBlockNumber < block.number;
+    function addressNotFrozen(address _owner) internal view returns (bool) {
```
zelig, line 221: [CRITICAL] this will allow the same node address to run several (any number of) nodes on different overlays and act as if staked on all of them!!

If you do it this way, overlay should also be passed from the redis contract and checked against the current one  recorded in the stakes map.

src/Staking.sol
```diff
@@ -196,45 +213,45 @@ contract StakeRegistry is AccessControl, Pausable {
...
 
     /**
-     * @dev Returns the current `stakeAmount` of `overlay`.
-     * @param overlay Overlay of node
+     * @dev Returns the current `stakeAmount` of `address`.
+     * @param _owner _owner of node
      */
-    function stakeOfOverlay(bytes32 overlay) public view returns (uint256) {
-        return stakes[overlay].stakeAmount;
+    function stakeOfAddress(address _owner) public view returns (uint256) {
```
zelig, line 229: [CRITICAL] see previous comment

src/Staking.sol
```diff
@@ -196,45 +213,45 @@ contract StakeRegistry is AccessControl, Pausable {
...
     /**
-     * @dev Returns the current usable `stakeAmount` of `overlay`.
+     * @dev Returns the current usable `stakeAmount` of `address`.
      * Checks whether the stake is currently frozen.
-     * @param overlay Overlay of node
+     * @param _owner owner of node
      */
-    function usableStakeOfOverlay(bytes32 overlay) public view returns (uint256) {
-        return overlayNotFrozen(overlay) ? stakes[overlay].stakeAmount : 0;
+    function usableStakeOfAddress(address _owner) public view returns (uint256) {
```
zelig, line 238: [CRITICAL] see previous comment

src/Staking.sol
```diff
@@ -196,45 +213,45 @@ contract StakeRegistry is AccessControl, Pausable {
...
+        return addressNotFrozen(_owner) ? stakes[_owner].stakeAmount : 0;
     }
 
     /**
-     * @dev Returns the `lastUpdatedBlockNumber` of `overlay`.
+     * @dev Returns the `lastUpdatedBlockNumber` of `address`.
      */
-    function lastUpdatedBlockNumberOfOverlay(bytes32 overlay) public view returns (uint256) {
-        return stakes[overlay].lastUpdatedBlockNumber;
+    function lastUpdatedBlockNumberOfOverlay(address _owner) public view returns (uint256) {
```
zelig, line 245: if this is ever called on its own, it should check the overlay too. 
If not, then it implies two map lookups and should be considered spurious gas spending.

In fact I think this check should just be part of the useablestake call and be checked internally not externally

zelig (changes requested):
fundamental issues with how stake is queried, in its current form allows for a critical exploit: it is possible to play with several overlays derived from the same address simultaneously while using only one stake. This is easily fixable by passing the overlay as well as an argument (to functions called from redis contract to validate stake of a node) and checking the overlay against the one recorded to the address currently.

The chosen implementation (see discussion on research internal) is probably the best option but it requires extensive changes in the interfaces therefore the most work for the bee team.  Also it needs a thorough change to the SWIP-19

src/Staking.sol
```diff
@@ -48,17 +45,22 @@ contract StakeRegistry is AccessControl, Pausable {
...
-     * @dev Emitted when a stake for overlay `frozen` for `time` blocks.
+     * @dev Emitted when a stake for address `frozen` is frozen for `time` blocks.
      */
-    event StakeFrozen(bytes32 slashed, uint256 time);
+    event StakeFrozen(address frozen, bytes32 overlay, uint256 time);
+
+    /**
+     * @dev Emitted when a address changes overlay it uses
+     */
+    event ChangedOverlay(address owner, bytes32 overlay);
```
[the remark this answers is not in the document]
zelig: should this not be indexed by address?

hardhat.config.ts
```diff
@@ -228,7 +228,7 @@ const config: HardhatUserConfig = {
     runOnCompile: false,
   },
   gasReporter: {
-    enabled: true,
```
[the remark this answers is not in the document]
n00b21337: this is just for gas checking while developing, will remove before merging.

src/Redistribution.sol
```diff
@@ -12,13 +12,13 @@ interface IPriceOracle {
 }
 
 interface IStakeRegistry {
-    function freezeDeposit(bytes32 overlay, uint256 time) external;
+    function freezeDeposit(address _owner, uint256 _time) external;
```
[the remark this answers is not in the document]
n00b21337: well I tried to make it minimal, so mostly changes are in staking which makes sense as we are changing what is key to mapping. bee node will mostly need to change interface on staking and not on redistribution.
It wouldn't make sense to have two mappings of stake data, that would be wasteful storage.

src/Redistribution.sol
```diff
@@ -300,11 +300,11 @@ contract Redistribution is AccessControl, Pausable {
             revert BelowMinimumStake();
         }
 
-        if (Stakes.ownerOfOverlay(_overlay) != msg.sender) {
+        if (Stakes.overlayOfAddress(msg.sender) != _overlay) {
```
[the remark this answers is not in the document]
n00b21337: the idea was that bee nodes don't need to change redistribution calls and that things stay the same as before with Redis ABI.
I do agree we can remove the overlay and get the value of it from stakes, that would be all right

src/Redistribution.sol
```diff
@@ -793,23 +793,23 @@ contract Redistribution is AccessControl, Pausable {
...
             revert MustStake2Rounds();
         }
 
-        if (Stakes.stakeOfOverlay(overlay) < MIN_STAKE) {
+        if (Stakes.stakeOfAddress(_owner) < MIN_STAKE) {
             revert BelowMinimumStake();
         }
 
-        return inProximity(overlay, currentRoundAnchor(), depth);
+        return inProximity(Stakes.overlayOfAddress(_owner), currentRoundAnchor(), _depth);
```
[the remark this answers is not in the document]
n00b21337: but they are not the same on is lookup for overlay and the other one is lookup for stake amount?

image

src/Staking.sol
```diff
@@ -48,17 +45,22 @@ contract StakeRegistry is AccessControl, Pausable {
...
      */
-    event StakeSlashed(bytes32 slashed, uint256 amount);
+    event StakeSlashed(address slashed, bytes32 overlay, uint256 amount);
 
     /**
-     * @dev Emitted when a stake for overlay `frozen` for `time` blocks.
+     * @dev Emitted when a stake for address `frozen` is frozen for `time` blocks.
      */
-    event StakeFrozen(bytes32 slashed, uint256 time);
+    event StakeFrozen(address frozen, bytes32 overlay, uint256 time);
```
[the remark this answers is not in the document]
n00b21337: same as above

src/Staking.sol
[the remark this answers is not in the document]
n00b21337: same as above, basically it comes down to this. a bit of different JS code but not much and maybe gas savings is worth it not to have indexed all around

Indexed
image

Not indexed
image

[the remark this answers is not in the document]
n00b21337: ok changing it

src/Staking.sol
```diff
@@ -89,84 +91,99 @@ contract StakeRegistry is AccessControl, Pausable {
...
-        emit StakeUpdated(overlay, updatedAmount, _owner, block.number);
+        emit StakeUpdated(_owner, updatedAmount, overlay, block.number);
     }
 
     /**
      * @dev Withdraw stake only when the staking contract is paused,
      * can only be called by the owner specific to the associated `overlay`
-     * @param overlay The overlay to withdraw from
-     * @param amount The amount of ERC20 tokens to be withdrawn
+     * @param _owner The _owner address to withdraw from
```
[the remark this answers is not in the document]
n00b21337: yeah, makes sense, was modeled per the previous WD for overlay but no need to have that extra parameters and lines of code.

src/Staking.sol
n00b21337: yeah, good call. changing this

src/Staking.sol
n00b21337: Copying this from private discussion

I dont think there is an attack vector. 
There is no case where I see that node can have 2 overlays staked
As we have multiple safe guards
There is only one overlay associated per address mapping at the time
Also when overlay changes, then we have lastUpdatedBlockNumber reset, which forces nodes out of the game for 2 rounds.
We have overlay checked on commits, so what nodes submit is checked per what is in staking mapping and needs to be the same, if not, the node cant submit or play. 
As we are using msg.sender it cant be manipulated to submit for someone else

src/Staking.sol
```diff
@@ -89,84 +91,99 @@ contract StakeRegistry is AccessControl, Pausable {
...
-        emit StakeSlashed(overlay, amount);
+        emit StakeSlashed(_owner, stakes[_owner].overlay, _amount);
+    }
+
+    /**
+     * @dev Change overlay of address to new one for neighbourhood hopping
+     * @param _owner the _owner adress selected
+     * @param _nonce the new nonce that will produce overlay
+     */
+    function changeOverlay(address _owner, bytes32 _nonce) external {
```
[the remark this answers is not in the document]
n00b21337: technically yes, that is kind of cool way to combine two and get results,
but then again we also have in depositStake **amount** so we have to add extra IF for overlay changes not to call token transfer on 0 and I am more leaning into separation of concerns and to have more readable code in this way. It would be a bit more gas spending on deployment which is fine and later it would be a little less gas for change overlay function and also we can have clean events for each action.

src/Staking.sol
```diff
@@ -48,17 +45,22 @@ contract StakeRegistry is AccessControl, Pausable {
...
      */
-    event StakeUpdated(bytes32 indexed overlay, uint256 stakeAmount, address owner, uint256 lastUpdatedBlock);
+    event StakeUpdated(address indexed owner, uint256 stakeAmount, bytes32 overlay, uint256 lastUpdatedBlock);
 
     /**
-     * @dev Emitted when a stake for overlay `slashed` is slashed by `amount`.
+     * @dev Emitted when a stake for address `slashed` is slashed by `amount`.
      */
-    event StakeSlashed(bytes32 slashed, uint256 amount);
+    event StakeSlashed(address slashed, bytes32 overlay, uint256 amount);
```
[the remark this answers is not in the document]
janos: From the SwarmScan perspective, indexed parameters are not needed as we can index them internally. It is not filtering events by parameters. But, if some other tools need filtering directly on the blockchain, indexing parameters is fine, of course.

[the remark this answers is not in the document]
n00b21337: I dont know that there is a need, if we get to it that there is will add it, thnx for info.

src/Staking.sol
n00b21337: hmm should be called ofAddress, also its read function that is useful for the debugging state of stake.
In redis could be called as part of reading of address mapping and comparing values.

src/Staking.sol
zelig: indeed no attack, but it has nothing to do with whats in the staking contract.

it is solely because this line in redis
https://github.com/ethersphere/storage-incentives/blob/25e7b18d6140a8290b905c2c76bbbaad94d23038/src/Redistribution.sol#L303
and this
https://github.com/ethersphere/storage-incentives/blob/25e7b18d6140a8290b905c2c76bbbaad94d23038/src/Redistribution.sol#L375

zelig (changes requested):
i took back the attack story, the overlay is checked here https://github.com/ethersphere/storage-incentives/blob/25e7b18d6140a8290b905c2c76bbbaad94d23038/src/Redistribution.sol#L303

but still issues left

src/Redistribution.sol
```diff
@@ -793,23 +793,23 @@ contract Redistribution is AccessControl, Pausable {
 
     /**
      * @notice Determine if a the owner of a given overlay can participate in the upcoming round.
-     * @param overlay The overlay address of the applicant.
-     * @param depth The storage depth the applicant intends to report.
+     * @param _owner The address of the applicant from.
```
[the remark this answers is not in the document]
n00b21337: Actually they can't this is just read function to check if node will be able to participate in next round, could be changed to msg.sender but it has also a function for manual checking/debugging of nodes and are they playing or not.  I can do 2 overloading functions here for ease of use.

src/Redistribution.sol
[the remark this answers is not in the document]
n00b21337: implemented

src/Staking.sol
```diff
@@ -234,7 +234,7 @@ contract StakeRegistry is AccessControl, Pausable {
     /**
      * @dev Returns the `lastUpdatedBlockNumber` of `address`.
      */
-    function lastUpdatedBlockNumberOfOverlay(address _owner) public view returns (uint256) {
+    function lastUpdatedBlockNumberOfAddress(address _owner) public view returns (uint256) {
```
zelig, line 237: has this not become unused as a result of getStake?

zelig (approved):

src/Redistribution.sol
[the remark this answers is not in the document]
nugaon: Either keep only this function and nodes will invoke it with their own address or create a wrapper that overloads this function without the `_owner` param and calls this function with `msg.sender`. Logic duplication is not a great practice.

I would choose the first option.

nugaon (changes requested):
Nothing critical but it would make difficulties with maintainability in a long run and confusion around keeping track overlay changes.

Also check the linter errors, the code has many because of the `max-line-length` rule which is set to 120.

zelig (approved):
I second @nugaon 's comments
and wonder why you removed the stake struct lookup to prevent repeated map lookups

src/Redistribution.sol
```diff
@@ -12,13 +12,22 @@ interface IPriceOracle {
...
-    function lastUpdatedBlockNumberOfOverlay(bytes32 overlay) external view returns (uint256);
+    function lastUpdatedBlockNumberOfAddress(address _owner) external view returns (uint256);
 
-    function ownerOfOverlay(bytes32 overlay) external view returns (address);
+    function overlayOfAddress(address _owner) external view returns (bytes32);
 
-    function stakeOfOverlay(bytes32 overlay) external view returns (uint256);
+    function stakeOfAddress(address _owner) external view returns (uint256);
+
+    function getStakeStruct(address _owner) external view returns (Stake memory);
```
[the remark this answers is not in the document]
nugaon: or is it?

ldeffenb:
First, is there going to be a migration of the existing stakes from overlay-based to owner-based?

And if so, has anyone given thought (or better yet, a test) to what happens with multiple abandoned overlay-based stakes that are all owned by a single account?   This happens when people keep their keys, but delete the rest of their data-dir causing a new overlay to be generated from the original account.

See this (short) conversation in Discord.  Start at the link and follow the replies backwards in time.
https://discord.com/channels/799027393297514537/811553590170353685/1251173722245693460

n00b21337:
> First, is there going to be a migration of the existing stakes from overlay-based to owner-based?
> 
> And if so, has anyone given thought (or better yet, a test) to what happens with multiple abandoned overlay-based stakes that are all owned by a single account? This happens when people keep their keys, but delete the rest of their data-dir causing a new overlay to be generated from the original account.
> 
> See this (short) conversation in Discord. Start at the link and follow the replies backwards in time. https://discord.com/channels/799027393297514537/811553590170353685/1251173722245693460

yes, there will be migration but basically it will be from users/operators side and we will need to write instructions on that and communicate them. Basically old contract would be paused, nodes would be taking funds out and could put them in new contract for staking.

About abandoned overlays, if the user has private key of the address used for that node he should be able to take funds out, if he knows the overlay, he can just input it and take funds out. 
If ppl are forgetting what they own, we could take data from blockchain and make a document with which address has which overlay and what amount so they could search for it there and take the funds out for themselves.

nugaon (approved):

PR #260, opened 2024-05-25
fix: rempove isValue as it is not needed

n00b21337:
We are using isValue in struct as a check if there is some value in struct, but it's obsolete, as solidity, has default values for values that are not set, so we can use "lastUpdatedBlockNumber" check to see if it is zero, as it will be zero just when staking is not ever added, the first time it is, it will have value and will always have some value in future, which is what we wanted with isValue in first place.


nugaon (approved):
maybe add comment to `lastUpdatedBlockNumber` is used for dirty flag as well

PR #263, opened 2024-05-28
fix: Remove pauser role

n00b21337:
As we currently use the same multistig for pauser and for default admin there is probably no need to have both roles defined and set and then used from the same address. So I am proposing we remove pauser role and just use admin role.


nugaon (approved):
if it is not used (or not planned in any way) then I also do not see any problem to remove it.

n00b21337:
> if it is not used (or not planned in any way) then I also do not see any problem with removing it.

It would make sense if we have one set of ppl that use multisig for pausing and the other set for general admin with different multisig. Was hard to even get ppl just for admin and to sign things, don't see we will be doing this kind of separation.

PR #275, opened 2024-09-24
feat: SWIP 21

n00b21337:
Implementation of SWIP on how to extend node storage capacity dedicated to the reserve to be able to calibrate operators' profitability.

More detail on this in SWIP writeup
https://github.com/ethersphere/SWIPs/pull/56


test/Redistribution.test.ts
```diff
@@ -93,11 +98,13 @@ const nonce_3 = '0xb5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555
 const hash_3 = '0xb5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33';
 const depth_3 = '0x06';
 const reveal_nonce_3 = '0xb5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33b5555b33';
+const height_3 = 0;
```
nugaon, line 101: not used.

nugaon (changes requested):
I tried to find some other occurrence when `height` is handled outside of `calculateEffectiveStake` but I did not find. 
I thought that `manageStake` will do something with the stake based on the height value.

nugaon:
also, please add a test when a node could not play in the round with height 0

src/Staking.sol
```diff
@@ -277,12 +294,21 @@ contract StakeRegistry is AccessControl, Pausable {
...
+
     function calculateEffectiveStake(
         uint256 committedStake,
-        uint256 potentialStakeBalance
+        uint256 potentialStakeBalance,
+        uint8 height
     ) internal view returns (uint256) {
         // Calculate the product of committedStake and unitPrice to get price in BZZ
-        uint256 committedStakeBzz = committedStake * OracleContract.currentPrice();
+        uint256 committedStakeBzz = (2 ** height) * committedStake * OracleContract.currentPrice();
```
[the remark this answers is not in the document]
n00b21337: not,it will be lower in stakeDenisty here because of height

```
uint8 depthResponsibility = _depth - revealedCommit.height;
stakeDensity: revealedCommit.stake * uint256(2 ** depthResponsibility),
```

also even if height  is multiplier on first line, finely value used will be lower one as (min) is used so doesn't matter much

```
       uint256 committedStakeBzz = (2 ** height) * committedStake * OracleContract.currentPrice();

        // Return the minimum value between committedStakeBzz and potentialStakeBalance
        if (committedStakeBzz < potentialStakeBalance) {
            return committedStakeBzz;
        } else {
            return potentialStakeBalance;
        }
```

n00b21337:
> also, please add a test when a node could not play in the round with height 0



> also, please add a test when a node could not play in the round with height 0

I tried to make something like this, The problem is that when you change height you are triggering "MustStake2Rounds" as it is affecting the staking struct and we require by default that any staking changes enforces node out of play for 2 rounds. which is not a problem really as it wont be playing in those 2 rounds anyway but its problem for testing as actually after 2 rounds anchor is different so we are not testing much other then if node can play on different anchor with higher height which we already did test in other test. For node not playing with height 0 we already have test from before where we didn't even have height and we just added default one now which is zero.

src/Redistribution.sol
```diff
@@ -372,9 +377,10 @@ contract Redistribution is AccessControl, Pausable {
         bytes32 obfuscatedHash = wrapCommit(_overlay, _depth, _hash, _revealNonce);
         uint256 id = findCommit(_overlay, obfuscatedHash);
         Commit memory revealedCommit = currentCommits[id];
+        uint8 depthResponsibility = _depth - revealedCommit.height;
```
istae, line 380: there is no need to subtract here. we should use _depth directly. 
bee will submit it's current storage radius, which will have height already subtracted from it.

src/Redistribution.sol
```diff
@@ -802,6 +808,7 @@ contract Redistribution is AccessControl, Pausable {
      */
     function isParticipatingInUpcomingRound(address _owner, uint8 _depth) public view returns (bool) {
         uint256 _lastUpdate = Stakes.lastUpdatedBlockNumberOfAddress(_owner);
+        uint8 _depthResponsibility = _depth - Stakes.heightOfAddress(_owner);
```
istae, line 811: again, depth is sufficient here

nugaon:
> I tried to make something like this, The problem is that when you change height you are triggering "MustStake2Rounds" as it is affecting the staking struct and we require by default that any staking changes enforces node out of play for 2 rounds

Mine blocks until you get an anchor that satisfies being out of the node's proximity with the storageDepth, but it can play with height > 0.

Issue #166, opened 2023-07-15
feat: add signer / 'delegate' field


mfw78:
**Problem**

Refer to the postage batch struct: https://github.com/ethersphere/storage-incentives/blob/e97ec1033843dbd373857597332e43dfc209806c/src/PostageStamp.sol#L61-L74

This provides pretty average UX / DX in that the `owner` **MUST** always be the same as the signer for stamps applied to chunks. This is a large handbrake on some particularly important features that would greatly smooth UX / DX.

**Solution**

Considering that stamps that have been processed by Swarm clients are in effect immutable, I propose:

1. The `owner` field is changed to a `signer` field. This represents the public address of the signer that must sign the chunks, as inspected by Swarm clients when exchanging data over Swarm. The `signer` would still effectively have root control over the batch.
2. An `owner` field is created, which can be instantiated on `createBatch` to a custom address. This `owner` has the additional benefits of being able to call dilute the batch and set a `delegate` as described below.
3. A `delegate` field (or better terminology TBD), which is configurable on `createBatch`, and configurable by `owner`. This field can also call the `topUp` and `dilute` functions for the respective batch.

**Use cases**

This allows for atomic dilution of a batch to increase it's capacity. Mandating that `dilute` can only be called by the `owner` requires the `owner` to posses gas, which prohibits smart contracts / smart contract wallets from possessing `batches`, which is prohibitively exclusive given the moves towards account abstraction.

mfw78:
Actually, a much tighter solution to this would be to enable an EIP-712 signatures on `dilute` so that `owner` could sign a dilute. This would represent significant UX/DX experience, allowing for many dilutes to be batched effectively, and atomically.

PR #282, opened 2025-01-31
feat: change stamp to fit for multichain solution

0xCardinalError:
To make postagestamp work properly with multichain and that we can distinguish who paid stamps and who created it we added creator value to batch. Creator will be Node address that is created stamp and uploading data, but owner may be not but could also be a user that is just uploading data through multichain solution dapp or some other way. So we can distinguish who is paying/owning the stamp and who is creating and uploading.

Will require migration of current stamps 
Will require bee node to change behaviour as there is more parameters and also batch ID is created per owner address and not msg.sender.
Once this is implemented, then multichain solution can created batchIDs based on owner of stamp and upload as such.
The same thing needs to be applied to bee nodes, that upload header for "swarm-postage-batch-id" is matching owner plus nonce.

### DIFF
--- a/src/PostageStamp.sol
+++ b/src/PostageStamp.sol
@@ -218,7 +218,7 @@ contract PostageStamp is AccessControl, Pausable {
         }
 
         uint256 totalAmount = _initialBalancePerChunk * (1 << _depth);
-        // We keep msg.sender as tokens can be transfered through smart contract from the owner
+        // We keep msg.sender as tokens can be transfered through proxy such as LIFI smart contract from the owner
         if (!ERC20(bzzToken).transferFrom(msg.sender, address(this), totalAmount)) {
             revert TransferFailed();
         }
@@ -389,7 +389,7 @@ contract PostageStamp is AccessControl, Pausable {
     function increaseDepth(bytes32 _batchId, uint8 _newDepth) external whenNotPaused {
         Batch memory batch = batches[_batchId];
 
-        if (batch.owner != msg.sender) {
+        if (batch.creator != msg.sender) {
             revert NotBatchOwner();
         }
 
--- a/src/PostageStamp.sol
+++ b/src/PostageStamp.sol
@@ -75,6 +75,8 @@ contract PostageStamp is AccessControl, Pausable {
     struct Batch {
         // Owner of this batch (0 if not valid).
         address owner;
+        // Creator of the batch, node that created the batch.
+        address creator;
         // Current depth of this batch.
         uint8 depth;
         // Bucket depth defined in this batch
@@ -90,6 +92,7 @@ contract PostageStamp is AccessControl, Pausable {
     struct ImportBatch {
         bytes32 batchId;
         address owner;
+        address creator;
         uint8 depth;
         uint8 bucketDepth;
         bool immutableFlag;
@@ -106,6 +109,7 @@ contract PostageStamp is AccessControl, Pausable {
         uint256 totalAmount,
         uint256 normalisedBalance,
         address owner,
+        address creator,
         uint8 depth,
         uint8 bucketDepth,
         bool immutableFlag
@@ -180,7 +184,8 @@ contract PostageStamp is AccessControl, Pausable {
     /**
      * @notice Create a new batch.
      * @dev At least `_initialBalancePerChunk*2^depth` tokens must be approved in the ERC20 token contract.
-     * @param _owner Owner of the new batch.
+     * @param _owner Owner of the new batch, can be same as creator if node is owner
+     * @param _creator Node that created the new batch.
      * @param _initialBalancePerChunk Initial balance per chunk.
      * @param _depth Initial depth of the new batch.
      * @param _nonce A random value used in the batch id derivation to allow multiple batches per owner.
@@ -188,6 +193,7 @@ contract PostageStamp is AccessControl, Pausable {
      */
     function createBatch(
         address _owner,
+        address _creator,
         uint256 _initialBalancePerChunk,
         uint8 _depth,
         uint8 _bucketDepth,
@@ -202,7 +208,7 @@ contract PostageStamp is AccessControl, Pausable {
             revert InvalidDepth();
         }
 
-        bytes32 batchId = keccak256(abi.encode(msg.sender, _nonce));
+        bytes32 batchId = keccak256(abi.encode(_owner, _nonce));
         if (batches[batchId].owner != address(0)) {
             revert BatchExists();
         }
@@ -212,6 +218,7 @@ contract PostageStamp is AccessControl, Pausable {
         }
 
         uint256 totalAmount = _initialBalancePerChunk * (1 << _depth);
+        // We keep msg.sender as tokens can be transfered through smart contract from the owner
         if (!ERC20(bzzToken).transferFrom(msg.sender, address(this), totalAmount)) {
             revert TransferFailed();
         }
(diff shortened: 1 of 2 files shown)


mfw78:
In general, while this solution would work, it is fragile and requires changes to many components within the core Swarm tech stack. https://github.com/ethersphere/storage-incentives/issues/166#issuecomment-1636723382 was raised as an alternative that would support something like this.

If EIP-712 signatures were used, in combination with a timeout to prevent replay, this would allow for all "authed" functions that currently work only for `msg.sender` to be done by signature.

In which case, the whole abi for _existing_ functions could remain the same and then insert `bySig` equivalents that allow the passing of the pre-image along with a signature for recovery. The recovered address could subsequently act as the equivalent of `msg.sender`.

Given that this requires a smart contract change either way, I would think the extended solution of EIP-712 signatures would be more robust and provide for future use cases / extensibility without the constant need for smart contract migrations which is itself fragile as well.

0xCardinalError:
> In general, while this solution would work, it is fragile and requires changes to many components within the core Swarm tech stack. [#166 (comment)](https://github.com/ethersphere/storage-incentives/issues/166#issuecomment-1636723382) was raised as an alternative that would support something like this.
> 
> If EIP-712 signatures were used, in combination with a timeout to prevent replay, this would allow for all "authed" functions that currently work only for `msg.sender` to be done by signature.
> 
> In which case, the whole abi for _existing_ functions could remain the same and then insert `bySig` equivalents that allow the passing of the pre-image along with a signature for recovery. The recovered address could subsequently act as the equivalent of `msg.sender`.
> 
> Given that this requires a smart contract change either way, I would think the extended solution of EIP-712 signatures would be more robust and provide for future use cases / extensibility without the constant need for smart contract migrations which is itself fragile as well.

I remember you mentioning this.  I am all for including as many changes and features as we need in this so we don't need to change postagestamp again. 
Smart contracts wallets are coming and will be more dominant for sure. Petra upgrade will come soon and this EIP would be new standard https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md

For multichain and generally enabling feature that data can be uploaded with other addresses than the node we should have another value saved to blockchain as we want not only to enable any address to create stamp on any node(that provides this endpoint) but also to upload more data later at any point if stamp is not full. So I am thinking in line of having value saved in this struct for each batch.

I don't see here much of a change to nodes, they will still upload data with the "creator" or "node" value as now. But owner or payer data will be saved so the middle layer code can check for node is the uploader actually the owner of node and can he do that, then this middle layer would upload data as now over node.  Check the readme file  here
https://github.com/ethersphere/multichain
so to get maybe better idea what was idea behind this. We can also add delegate as another option if needed and dilution is again done through node as checker on smart contract but node can make a middle layer endpoint for doing that where it checks if the owner is the one making changes.

mfw78:
Fundamentally I don't understand why there needs to be a change to the core struct when all operations can be done by a signature, in which case there is no need to change any of the existing functions and the node implementations remain unaffected. 

Forget the delegates method or whatever. Simply all existing functions that require an owner as msg.sender could have signed analogue. This eliminates any need to touch the bee implementation whatsoever.

0xCardinalError:
Here are the reasons

1. We are using LIFI smart contracts to get stamps, cant implement sigs into their contracts, msg.sender is actually their smart contract
2. We can't reupload data to the same stamp if we don't have a record of who paid for it
3. We want to show users a dashboard of their stamps, again to do that, we need to have an easy fetchable record of their stamps.  Getting data through some logs is almost impossible to track as we are using cross chain swaps with various different protocols from LIFI aggregator, where there is few transactions happening and we would need to built a whole app of tracking trx origins that cover all the networks and protocols that might be used. This doesn't make sense if we can cover this with just one more value stored in struct.

mfw78:
If the LiFi smart contracts to get stamps allow for the passage of arbitrary calldata, then this allows LiFi to trigger different functions, other than `create` etc. In which case, one can pass calldata using `createBySig` which then has the values that are needed, as well as the `rsv` for a signature. Then by using `ecrecover` of the `rsv`, one can then determine which address it is that actually is the true `owner` of the stamp.

This still allows for the same semantics when purchasing stamps from cross-chain, where the app can maintain the current signer + the `nonce` that has been used, and in which case, is able to then determine the `batch_id` and query the information from the Gnosis Chain public RPC with minimal RPC calls.

Looking at https://docs.li.fi/li.fi-api/li.fi-api/requesting-a-quote/cross-chain-contract-calls (assuming this is what will be used) - it permits specifying the `callData`, so it's perfectly possible to pre-package the entire calldata, including the `rsv` for a signature that allows for passing everything across chains - and avoids changing the core struct which would otherwise then require client changes.

0xCardinalError:
Yes, signature could be used with calldata. But still then the ownership data is stored in trx, which needs to be searched for each time and recovered which I don't see viable as you cant search by user transaction, how would you pick that trx when its not connected directly to user account for data in use case 3 as it would originate from LIFI contract

What are we even trying to do here, make less changes to bee node?! I don't see this as the proper way to architect. 
I want to pick up the best solutions, not the ones with the least changes on the bee node. We still would need to make changes to the bee node for the new ABI function, stamps will need to be migrated either way. 
What you propose wouldn't change batch struct that is true.
What is the main goal of the way you propose changes, what do we want to achieve?

mfw78:
I'm a little confused by what you mean with regards to "ownership data is stored in trx". In `createBySig` batch.owner can be populated with the `ecrecover` address of the true owner, and then there is an event that is emitted for the batch being created that contains the owner?

The main goal that I seek is that doing things in this manner makes it so that it's straightforward with no changes required from a `bee` context (besides pointing to a new contract address) for dapps to be built that allow for cross-chain purchase of batches, and within the dapp themselves they can have their own wallet and make use of a local stamper, in combination with pre-signed chunks from the `bee` API endpoint (can push data to a local node - or a gateway etc).

When something can be architected to be narrow in scope and minimise / eliminate changes, I think this should be balanced / inform the appetite for risk associated with the change.

0xCardinalError:
Don't be confused :), it was late and I made an error in thinking about that part, you are right.
Will talk with Bee guys about the differences in implementation.

What is the idea behind sig?  usually, it is to avoid trx, but in this case, we need to make trx anyway because of swaps and stamp creation.

mfw78:
> What is the idea behind sig? usually, it is to avoid trx, but in this case, we need to make trx anyway because of swaps and stamp creation.

The idea behind the sig is so that _who executes the transaction_ becomes irrelevant. It can be done by LiFi, or as an example, it opens up Hook integrations with CoW Protocol that would allow async swaps to batch top-up etc. It's wildly more flexible.

Also, as part of this, what would be very handy would be if the batch ID could be refined as `H(owner|nonce)`, instead of it being done by `H(msg.sender|nonce)`. :grin:

0xCardinalError:
The main problem I see with not changing the struct is that we would need to change bee nodes more. 
As bee nodes will let uploads only for the batches they own, and to own them we need to have the batch.owner match the node.  You proposed to have it extracted from sig and match the owner who paid the batch and this is then non-uploadable when it hits the node

That is why I am suggesting a dual system, where basically things for node stays the same, we just change the var name owner => uploader  and add new "owner" var that is not even a concern to node code, but to its middleware that can and might not use depending on the node, most won't use this and would operate as before.

@istae can you weigh in from your perspective what do you think

mfw78:
https://docs.ethswarm.org/api/#tag/Chunk/paths/~1chunks/post

This method allows upload of chunks with a precalculated stamp (IE. Whereby the chunk is signed by an account that may differ to the node's).

0xCardinalError:
We wont use this as we are using proxy contract instead

PR #289, opened 2025-06-26
Fix:change order errors

0xCardiE:
This optimization will:
✅ Save gas by doing cheap checks first
✅ **Provide better user experience with logical error ordering**
✅ Exit early on the most common validation failures
✅ Only perform expensive modulo calculations when really needed


nugaon (approved):
good point to change it!

PR #292, opened 2025-06-28
fix: Fix racing condition in 2 player test

0xCardiE:
# Fix Intermittent CI Test Failure in Stats Test

## Problem

The "two player game is fair with 1:3 stake" test was failing intermittently in CI with the error:

This failure occurred randomly - sometimes the first run would fail, but rerunning the CI would succeed.

## Root Cause Analysis

The issue was a **race condition in witness caching**:

1. **Anchor-Dependent Witnesses**: Witnesses are generated based on a specific `anchor` value (from `currentSeed()`)
2. **Naive Caching**: The caching system used simple filenames (`stats-000.json`, `stats-001.json`, etc.) without considering the anchor
3. **Stale Cache Reuse**: When tests ran with different anchors, they would load cached witnesses generated for previous anchors
4. **Validation Failure**: The `transformedAddress` values in cached witnesses were invalid for the new anchor, causing the proof validation to fail

### 1. Anchor-Aware Caching
Modified the witness filename to include the anchor hash:
```typescript
const anchorHash = hexlify(anchor).slice(2, 10); // Use first 8 chars of anchor hex
const uniqueSuffix = `${suffix}-${anchorHash}-d${depth}${socType ? '-soc' : ''}`;
```

### 2. Witness Validation
Added `validateWitnessesForAnchor()` function to verify cached witnesses are still valid:
- Checks that `transformedAddress` matches expected value for current anchor
- Validates witnesses satisfy depth requirements
- Regenerates witnesses if validation fails

### 3. Backward Compatibility
Maintained compatibility with existing cached witness files:
- First tries new naming pattern
- Falls back to old naming pattern if new files don't exist
- Validates old cached witnesses before using them

### 4. Cleanup Mechanism
Added automatic cleanup to prevent directory bloat:
- Keeps only the 50 most recent files per prefix
- Runs automatically when generating new witnesses

## Files Changed

- `test/util/proofs.ts`: Enhanced `setWitnesses()` function with validation, cleanup, and backward compatibility

## Benefits

✅ **Eliminates Race Condition**: Each unique anchor gets its own cached witnesses  
✅ **Maintains Performance**: Still uses caching when possible (test runs in ~11s vs several minutes)  
✅ **Self-Healing**: Automatically detects and fixes invalid cached witnesses  
✅ **Backward Compatible**: Existing cached witness files continue to work  
✅ **Prevents Bloat**: Automatic cleanup of old witness files  
✅ **Clean Output**: No verbose logging messages cluttering test output  

## Testing

The fix has been verified to:
- Use existing cached witnesses when they're valid for the current anchor
- Regenerate witnesses only when necessary
- Complete the test suite quickly (~11 seconds)
- Eliminate the intermittent "transformed witness chunk does not match" error

This should resolve the CI reliability issues while maintaining optimal performance.

### DIFF
--- a/test/util/chai.ts
+++ b/test/util/chai.ts
@@ -3,4 +3,5 @@ import { chaiEthers } from 'chai-ethers';
 
 chaiModule.use(chaiEthers);
 
-export = chaiModule;
+export const expect = chaiModule.expect;
+export default chaiModule;
--- a/test/util/proofs.ts
+++ b/test/util/proofs.ts
@@ -391,7 +391,6 @@ export function loadWitnesses(filename: string): WitnessData[] {
 }
 
 export function saveWitnesses(witnessChunks: WitnessData[], filename: string) {
-  console.log('save witnesses');
   fs.writeFileSync(
     path.join(__dirname, '..', 'mined-witnesses', `${filename}.json`),
     JSON.stringify(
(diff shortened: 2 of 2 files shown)


0xCardiE:
Actually very slow now and will come to this at some later time.

PR #301, opened 2026-01-12
Fix - remove skipped rounds guard

0xCardiE:


significance:
after some discussion it was understood this was in fact guarding against other angles of attack

PR #303, opened 2026-01-15
feat: defang skipped rounds penalty 

0xCardiE:
Changes the skipped rounds price adjustment to use changeRate[1] (as if 1 player participated) instead of changeRate[0] (maximum penalty for no players).


0xCardiE:
Difference between steps here

image


So this would mean we have now double +1 amount of value when there is 0 commit then when there is 1 and this change in PR would make it half of it.

zelig:
see https://github.com/ethersphere/research-tasks/issues/9

mfw78:
> see https://github.com/ethersphere/research-tasks/issues/9

Can't see this as the repo is private :smiling_face_with_tear:

PR #306, opened 2026-02-28
Add Echidna fuzz testing harness + runner 

0xCardiE:
Introduce Echidna-based, stateful fuzz testing for this repo.

Add a Docker-based runner and documentation, plus a multi-actor harness that fuzzes all smart contracts with role modeling, cross-actor accounting invariants, non-interference checks, and post-condition properties for all contracts.

**What is fuzz testing?**  
Fuzz testing ("fuzzing") is an automated testing technique that repeatedly calls contract functions with many randomized inputs and stateful sequences of calls. Instead of writing a single expected-output test per scenario, you write properties/invariants that must always hold, and the fuzzer tries hard to find a sequence that breaks them.

**Why use it for smart contracts?**  
Finds edge cases humans miss: weird input combinations, unusual call ordering, boundary values, and unexpected state transitions.  
Stateful bugs: catches issues that only appear after many steps (e.g. stake -> freeze -> slash -> migrate -> restake).  
Property-driven: you encode "this should never happen" rules (access control, accounting conservation, deletion semantics), and the tool searches for counterexamples.

### What we test

**Commit** (well covered)

- `act_happyCommit` in the redistribution harnesses constructs valid overlays, sets staking mock data, and calls `commit()` through actors
- Properties verify commits land in storage correctly (`echidna_tracked_commit_matches_storage`, `echidna_commit_overlays_unique`)

**Reveal** (covered, though hit rate is still probabilistic in the general harnesses)

- `act_happyReveal` replays tracked pre-images from the same round
- Properties verify reveal storage matches (`echidna_tracked_reveal_matches_storage`, `echidna_revealed_commit_indices_valid`, `echidna_reveal_entries_imply_matching_commit`)

**Winner selection** (covered via exposed internal)

- `act_winnerSelection` in the base harness calls `exposedWinnerSelection()` which runs the real `winnerSelection()` internal function - this is the core claim logic:
  - `getCurrentTruth()` truth determination
  - winner random selection weighted by stake density
  - freezing non-revealers and disagreeing revealers
  - `OracleContract.adjustPrice()` call
  - `currentClaimRound` bookkeeping
- Properties: `echidna_winnerSelection_only_once_per_round`, `echidna_last_winnerSelection_freezes_nonrevealed`

**Pot withdrawal** (covered via claim stub)

- `claimStub()` in the claim-stub harness runs real `winnerSelection()` and then calls `PostageStamp.withdraw(winner)` directly
- Properties: `echidna_claim_withdraws_pot_to_winner_when_successful`, `echidna_claim_triggers_oracle_adjustPrice`, `echidna_claim_only_once_per_round`

**Real `claim()` proof path** (now covered with fixture-based fuzzing)

- `EchidnaRedistributionRealClaimHarness` exercises the real `Redistribution.claim()` verifier path with fixed valid CAC and SOC proof bundles
- The harness prepares a real `commit -> reveal -> claim` flow, then runs the actual proof-verifying `claim()` function
- Instead of trying to randomly discover valid cryptographic proofs, it starts from known-good fixtures and fuzzes targeted mutations around them
- Mutations cover selected inclusion-proof fields, postage index fields, and SOC fields while keeping the rest of the scenario realistic
- Properties:
  - `echidna_unmutated_fixture_claim_succeeds`
  - `echidna_mutated_fixture_claim_does_not_succeed`
  - `echidna_successful_real_claim_effects_hold`

### Why the real `claim()` path needs a fixture-based approach

Naive random fuzzing is very unlikely to hit a valid proof-verifying `claim()` because the inputs must satisfy multiple correlated cryptographic relationships at once:

- correct BMT Merkle inclusion proofs
- valid postage signatures for the batch owner
- valid SOC signatures and address derivation
- matching proximity relationships against the round anchor
- correct ordering and reserve-estimation conditions

These witnesses need to be constructed, not discovered by chance. The fixture-based harness solves this by seeding Echidna with valid proof bundles and then fuzzing mutations around them while still executing the real on-chain verifier logic.

### Is the current coverage sufficient?

The split is now more complete and more intentional:

- **Echidna covers**: the state machine (phase transitions, round bookkeeping, once-per-round constraints), economics (pot distribution, freezing, oracle updates), access control, and now the real `claim()` verifier path via fixed CAC/SOC fixtures plus targeted proof mutations
- **Chai unit tests still cover best**: targeted, readable proof-verification scenarios with hand-crafted valid/invalid proofs and exact expected revert reasons

### Supporting runner/config changes

Also added:

- `echidna/echidna-real-claim.yaml`
- `ECHIDNA_CONFIG` support in `scripts/echidna.sh`

This lets the real-claim harness use its own isolated corpus and run independently from the shared suite when needed.

### Validation

Validated both with the dedicated run:

```bash
ECHIDNA_CONTRACT=EchidnaRedistributionRealClaimHarness \
ECHIDNA_CONFIG=echidna/echidna-real-claim.yaml \
yarn echidna

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -101,6 +101,29 @@ To get started with this project, follow these steps:
...
+
+```bash
+yarn echidna
+```
+
+### What this is doing
+
+Echidna repeatedly calls the harness “action” functions (like `act_manageStake`) with random inputs, building **stateful sequences**. After (and during) those sequences it checks `echidna_*` property functions such as:
+
+- **Token properties**: total supply stays constant; decimals stay 16
```
nugaon: I cannot imagine [the decimals](https://github.com/ethersphere/storage-incentives/blob/9694920eff72d7bf6ff15f162b54647b56a09c51/src/TestToken.sol#L17-L18) could change anyhow

README.md
```diff
@@ -101,6 +101,29 @@ To get started with this project, follow these steps:
 2. Run `yarn install` at the root of the repo to install all dependencies.
 3. Add a `.env` file in your root directory, where you'll store your sensitive information for deployment. An example file [`.env.example`](./.env.example) is provided for reference.
 
+## Fuzz testing (Echidna)
+
+This repo includes a **small Echidna harness** that fuzzes `TestToken` + `StakeRegistry` with a few basic invariants.
+
+- **Harness contract**: `src/echidna/EchidnaStakeRegistryHarness.sol`
+- **Echidna config**: `echidna/echidna.yaml`
```
nugaon: This section could just refer the echidna/README.md with minimal description to run echidna tests instead of elaborating the functionalities (which seems outdated anyway)

scripts/echidna.sh
```diff
@@ -0,0 +1,48 @@
...
+  CONTRACTS_TO_RUN=("$CONTRACT")
+else
+  CONTRACTS_TO_RUN=("${CONTRACTS_DEFAULT[@]}")
+fi
+
+for c in "${CONTRACTS_TO_RUN[@]}"; do
+  echo "==> echidna: running contract $c" >&2
+
+  # Avoid stale Crytic compile artifacts causing old properties/tests to run.
+  rm -rf crytic-export
```
nugaon: This is wrong, the docker writes the files with root, and my user permission cannot remove it.

src/echidna/EchidnaPriceOracleHarness.sol
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    function act_admin_setPrice(uint32 p) external {
+        _clearPending();
+
+        pendingStampCallsBefore = stamp.setPriceCalls();
+
+        uint64 minUp = oracle.minimumPriceUpscaled();
+        uint64 expected = uint64(p) << 10;
+        if (expected < minUp) expected = minUp;
```
nugaon: why do you override `expected` like this? seemingly, the bit overflow is surpressed like this, but it can hide other problems too.

src/echidna/EchidnaPriceOracleHarness.sol
```diff
@@ -0,0 +1,321 @@
...
+
+        pendingExpectedUpScaled = expected;
+        pendingStampShouldCall = !stamp.shouldRevert();
+        pendingSetPrice = true;
+
+        (bool ok, bool returned) = _callSetPriceAsAdmin(p);
+        ok;
+        returned;
```
nugaon: Simply just call `_callSetPriceAsAdmin(p);` - on other places as well. 
but the main issue is pendingSetPrice = true set before this call. If it fails, there is no pending price set happened here.

src/echidna/EchidnaPriceOracleHarness.sol
```diff
@@ -0,0 +1,321 @@
...
+    PriceOracle internal immutable oracle;
+
+    constructor(PriceOracle oracle_) {
+        oracle = oracle_;
+    }
+
+    function callSetPrice(uint32 p) external returns (bool ok, bool returned) {
+        bytes memory data;
+        (ok, data) = address(oracle).call(abi.encodeWithSelector(oracle.setPrice.selector, p));
+        returned = ok && data.length >= 32 ? abi.decode(data, (bool)) : false;
```
nugaon, line 32: data.length is exactly 32

src/echidna/EchidnaPriceOracleHarness.sol
```diff
@@ -0,0 +1,321 @@
...
+        (bool canCompute, uint64 expected) = _tryExpectedAdjustedPrice(
+            pendingPriceBefore,
+            redundancy,
+            currentRound,
+            pendingLastAdjustedBefore
+        );
+
+        if (!canCompute) {
+            if (ok && returned) adjustWouldOverflowButSucceeded = true;
+            returned;
```
nugaon: this is redundant here. and so in other places as well. (returned not used anywhere

src/echidna/EchidnaPriceOracleHarness.sol
```diff
@@ -0,0 +1,321 @@
...
+        uint64 currentRound = oracle.currentRound();
+        bool wouldRevertEarly = (redundancy == 0) || (currentRound <= pendingLastAdjustedBefore);
+
+        if (wouldRevertEarly) {
+            // If it unexpectedly succeeded, that's a bug in either model or contract.
+            if (ok) adjustWouldOverflowButSucceeded = true;
+            // A revert should not change state.
+            if (oracle.currentPriceUpScaled() != pendingPriceBefore) pausedAdjustChangedState = true;
+            if (oracle.lastAdjustedRound() != pendingLastAdjustedBefore) pausedAdjustChangedState = true;
+            if (stamp.setPriceCalls() != pendingAdjustStampCallsBefore) pausedAdjustChangedState = true;
```
nugaon: these conditions could be merged with `OR` indicating these are related checks setting the same flag.

src/echidna/EchidnaPriceOracleHarness.sol
```diff
@@ -0,0 +1,321 @@
...
+
+    bool internal pendingAdjust;
+    bool internal pendingAdjustPaused;
+    uint64 internal pendingPriceBefore;
+    uint64 internal pendingLastAdjustedBefore;
+    uint64 internal pendingExpectedLastAdjustedAfter;
+    uint64 internal pendingExpectedUpScaledAfter;
+    uint256 internal pendingAdjustStampCallsBefore;
+    bool internal pendingAdjustStampShouldCall;
+    bool internal adjustWouldOverflowButSucceeded;
```
nugaon, line 77: it is not cleared anywhere.

src/echidna/EchidnaPriceOracleHarness.sol
```diff
@@ -0,0 +1,321 @@
...
+    bool internal unauthorizedAdjustSucceeded;
+    bool internal pausedAdjustChangedState;
+
+    // Pending post-conditions (cleared on each action).
+    bool internal pendingSetPrice;
+    uint64 internal pendingExpectedUpScaled;
+    uint256 internal pendingStampCallsBefore;
+    bool internal pendingStampShouldCall;
+
+    bool internal pendingAdjust;
```
nugaon, line 69: only pendingSetPrice and pendingAdjust are cleared by `_clearPending`

src/echidna/EchidnaSystemHarness.sol
```diff
@@ -0,0 +1,396 @@
...
+        return token.balanceOf(address(stake)) >= sumPotential;
+    }
+
+    function echidna_stamp_totalPot_never_exceeds_token_balance() external view returns (bool) {
+        // `totalPot()` is `min(pot, tokenBalance)` but is not `view` (it calls `expireLimited`),
+        // so we assert the semantic relationship in a view-safe way.
+        uint256 bal = token.balanceOf(address(stamp));
+        uint256 p = stamp.pot();
+        uint256 totalPotView = p < bal ? p : bal;
+        return totalPotView <= bal;
```
nugaon: always true, does not do much.

src/echidna/EchidnaRedistributionHarness.sol
```diff
@@ -0,0 +1,834 @@
...
+    bool internal revealSucceededWhilePaused;
+    bool internal winnerSelectionSucceededTwiceSameRound;
+
+    // Pending winnerSelection postconditions.
+    bool internal pendingWinnerSelection;
+    uint64 internal pendingWinnerSelectionRound;
+    uint8 internal pendingWinnerSelectionLen;
+    address[25] internal pendingWSOwners;
+    bool[25] internal pendingWSRevealed;
+    uint256[25] internal pendingWSFreezeCountBefore;
```
nugaon, line 114: no _clearPending method calls in this contract, it would be nice to be uniform across the contracts.

nugaon (commented):
Really nice initiative with really a lot of changes.
It would have been better in several different PRs.

I checked the basic setup and the PriceOrcale harness mostly, but I peeked into other harness logics as well.
I stumbled upon [an issue](https://github.com/ethersphere/storage-incentives/pull/306/files/9694920eff72d7bf6ff15f162b54647b56a09c51#diff-2639b94e6c6eaf57c67761afc5672033e87141b14f747e756a480ab6946a824b) with running the test, see below.

0xCardiE:
> Really nice initiative with really a lot of changes. It would have been better in several different PRs.
> 
> I checked the basic setup and the PriceOrcale harness mostly, but I peeked into other harness logics as well. I stumbled upon [an issue](https://github.com/ethersphere/storage-incentives/pull/306/files/9694920eff72d7bf6ff15f162b54647b56a09c51#diff-2639b94e6c6eaf57c67761afc5672033e87141b14f747e756a480ab6946a824b) with running the test, see below.

yes you are right. I remove one part into another PR that should be separate
https://github.com/ethersphere/storage-incentives/pull/308/changes

README.md
0xCardiE: True, was aimed for robustness and if in some cases code changes that we check every minor detail.
Logic wise how its now this would never happen.

scripts/echidna.sh
0xCardiE: Current scripts/echidna.sh no longer removes anything on the host. 
It runs rm -rf crytic-export inside the same docker run as Echidna (sh -c 'rm -rf crytic-export && echidna-test …')

README.md
nugaon: these kind of tests should be chai manual tests. shouldn't echidna check only state-model correctness? 
IMO this does not need detailed testing because it doesn't replace targeted tests.

scripts/echidna.sh
nugaon: I don't completely understand why crytic-export generated in the first place if you automatically remove it in each iteration?

.gitignore
```diff
@@ -38,4 +38,10 @@ contractsInfo.json
...
 
 # Tenderly
-tenderly.log
\ No newline at end of file
+tenderly.log
+
+# Echidna fuzzing
+echidna/corpus/
+echidna/coverage/
+echidna/out/
```
nugaon: I just have corpus folder after running echidna. coverage folder is in it, out folder is missing.

scripts/echidna.sh
```diff
@@ -0,0 +1,46 @@
...
+cd "$ROOT_DIR"
+
+IMAGE="${ECHIDNA_IMAGE:-ghcr.io/crytic/echidna/echidna:latest}"
+CONTRACT="${ECHIDNA_CONTRACT:-}"
+
+# Compile on the host. The Echidna container image doesn't ship with Node/npx,
+# and without Hardhat artifacts CryticCompile will try (and fail) to run `npx hardhat compile`.
+yarn -s hardhat compile --force >/dev/null
+
+CONTRACTS_DEFAULT=(
```
nugaon: it could be dynamic instead of hardcoded.

echidna/README.md
```diff
@@ -0,0 +1,255 @@
...
+- `echidna/coverage/`
+- `crytic-export/` (Crytic export artifacts)
+
+These are ignored by git via `.gitignore`.
+
+## How to extend this
+
+Typical next steps:
+
+- Add another harness under `src/echidna/` for other protocol contracts.
```
nugaon: also needed to add to the bash contracts array.

README.md
0xCardiE: Yes you are right
The right mental model would be

Chai tests = targeted, readable, scenario-driven tests
Echidna = randomized exploration of both global invariants and per-call correctness across arbitrary state histories

scripts/echidna.sh
0xCardiE: It's generated as a side effect of echidna-test. Every invocation of Echidna triggers CryticCompile, which writes to crytic-export/. We can't prevent that; we can only clean it before the next iteration so it doesn't contaminate the next contract's run.

0xCardiE:
Made serious cleanups per suggestions above, keeping fuzz tests that makes most sense to be in Echidna.

.cursor/plans/security_findings_exploitability_8fd9957b.plan.md
```diff
@@ -0,0 +1,280 @@
+---
+name: Security Findings Exploitability
```
nugaon: it is not connected to the PR.

src/echidna/EchidnaRedistributionClaimHarness.sol
```diff
@@ -0,0 +1,334 @@
...
+        trackedRound[idx] = redist.currentRound();
+        trackedObfuscated[idx] = obf;
+        trackedHash[idx] = hash;
+        trackedNonce[idx] = nonce;
+        trackedDepth[idx] = depth;
+    }
+
+    function act_happyReveal(uint8 actorId) external {
+        _clearClaimPending();
+        if (!redist.currentPhaseReveal()) return;
```
nugaon, line 248: it returns always here

src/echidna/EchidnaRedistributionHarness.sol
```diff
@@ -0,0 +1,796 @@
...
+            if (!ok) break;
+            CommitView memory cv = abi.decode(data, (CommitView));
+            pendingWSOwners[i] = cv.owner;
+            pendingWSRevealed[i] = cv.revealed;
+            pendingWSFreezeCountBefore[i] = stakeMock.freezeCount(cv.owner);
+            pendingWinnerSelectionLen++;
+        }
+
+        bool okCall = actors[idx].callWinnerSelection();
+        if (!okCall) return;
```
nugaon, line 345: it always returns at [this point](https://github.com/ethersphere/storage-incentives/blob/1fe97a1f8281c0fee28071b082ed46177f7b71c1/src/Redistribution.sol#L497-L502)

src/echidna/EchidnaRedistributionHarness.sol
```diff
@@ -0,0 +1,796 @@
...
+        trackedStake[idx] = stake;
+        trackedReserveHash[idx] = reserveHash;
+        trackedNonce[idx] = nonce;
+        trackedObfuscated[idx] = obfuscated;
+    }
+
+    function act_happyReveal(uint8 actorId) external {
+        _clearWinnerSelectionPending();
+        if (redist.paused()) return;
+        if (!redist.currentPhaseReveal()) return;
```
nugaon, line 300: it always returns at this point

nugaon (commented):
according the redistribution tests coverage in the corpus, the reveal and the claim function logic is not even touched by the echidna tests.
I commented some of the lines where the workflow breaks.
The testcode does not bypass the block number related checks in these functions and the runs have really low chance to hit any of the phases of the redistribution game. 
maybe `maxBlockDelay` could be changed in the Echidna conf file to a lower number.

0xCardiE:
> according the redistribution tests coverage in the corpus, the reveal and the claim function logic is not even touched by the echidna tests. I commented some of the lines where the workflow breaks. The testcode does not bypass the block number related checks in these functions and the runs have really low chance to hit any of the phases of the redistribution game. maybe `maxBlockDelay` could be changed in the Echidna conf file to a lower number.

great point, this setting was wrong completely. thank you

0xCardiE:
Fixed previous discrepancies and comments and also added some fuzzing to claim function and separate config and corpus for it.

nugaon:
I checked the new EchidnaRedistributionRealClaimHarness.sol but it has just partial coverage seemingly, e.g. act_prepareFixtureReveal does not go further after the redist.currentPhaseReveal() check.

src/echidna/EchidnaRedistributionHarness.sol
```diff
@@ -0,0 +1,640 @@
...
+        a.callCommit(obfuscatedHash, rn);
+    }
+
+    function act_reveal(uint8 actorId, uint8 depth, bytes32 hash, bytes32 nonce) external {
+        _clearWinnerSelectionPending();
+        EchidnaRedistributionActor a = actors[uint256(actorId) % ACTOR_COUNT];
+        a.callReveal(uint8(depth % 32), hash, nonce);
+    }
+
+    function act_claim(uint8 actorId) external {
```
nugaon, line 219: never runs

nugaon (commented):
I still don't get good coverage on claim at EchidnaPostageStampHarness.

the tests should be simplified, there are two different specific redistribution tests but seemingly none of them get to the claim phase (mostly stops at winnerselection).
as we discussed this PR is too heavy, rereading 3.5k lines all the time is not the best method that we could do... maybe that's why I have the feeling about this approach overcomplicates the codebase.

The fixes on the contracts seem good, we could merge those in a different fix PR.

src/echidna/EchidnaRedistributionHarness.sol
0xCardiE: image

It runs, I have multiple echidna corpuses where it was run. Sharing just one here

[covered.1777581556.html](https://github.com/user-attachments/files/28072440/covered.1777581556.html)

0xCardiE:
> I still don't get good coverage on claim at EchidnaPostageStampHarness.
> 
> the tests should be simplified, there are two different specific redistribution tests but seemingly none of them get to the claim phase (mostly stops at winnerselection). as we discussed this PR is too heavy, rereading 3.5k lines all the time is not the best method that we could do... maybe that's why I have the feeling about this approach overcomplicates the codebase.
> 
> The fixes on the contracts seem good, we could merge those in a different fix PR.

Coverage is pretty good on PostageStamp, we call all the actions there and its getting to 80% so very small part is not called.

image


They get to claim phase. Maybe we need to talk about this how it should be read, seems like we are reading info differently

image

We have some Claim stubs and mocks so we can try and test parts of it. Redistribution is complex contract that is why we have 2 harnesses, in first we test commit/reveal, winner selection in second we have claim/withdraw testing.
Both dont have real inclusion proofs as this is hard to fuzz so we need to simulate parts of it.

Yes contract fixes should be applied with specific PRs when we deploy new contracts and will be pulled out of this PR.

0xCardiE:
> I still don't get good coverage on claim at EchidnaPostageStampHarness.
> 
> the tests should be simplified, there are two different specific redistribution tests but seemingly none of them get to the claim phase (mostly stops at winnerselection). as we discussed this PR is too heavy, rereading 3.5k lines all the time is not the best method that we could do... maybe that's why I have the feeling about this approach overcomplicates the codebase.
> 
> The fixes on the contracts seem good, we could merge those in a different fix PR.

About line numbers, I dont see it that way. There are tests, they should be verbose, not minimized. Dont see it add any complexity as its separate part of codebase not affecting actual contracts. They are made in a way that we cover all functions in contract, with properties that make sense to be checked. At best we can remove some properties that you might see as not needed. Otherwise think this is fine, it aims to run all the functions in contracts and exposes them to fuzzing best it can.

nugaon (approved):
As we discussed, the fuzzing logic with the coverage will be revised later, but overall the current state of tests and configuration seems fine. Configurations, fixes, test setups seem to be in place and correct.

PR #308, opened 2026-04-07
fix: stabilize Stats and Redistribution tests on CI

0xCardiE:
- Harden Stats proof test against flakiness (deterministic witness flow)
- Adjust Redistribution tests for stable CI behavior

Stats are often flaky and fail more then succed

image


0xCardiE:
Reviewed the code, nothing functional is changed, we just make testing more stable here.

PR #311, opened 2026-05-22
fix(oracle): clamp upscaled price to uint32-safe bounds

0xCardiE:
## Summary
- Cast to `uint64` before upscaling in `setPrice` to avoid overflow/truncation when shifting.
- Add `MAX_CURRENT_PRICE_UPSCALED` and `_clampPriceUpscaled` so `currentPrice()` cannot disagree with `currentPriceUpScaled` or under-report vs `minimumPrice()`.

## Test plan
- [ ] Run existing PriceOracle unit tests
- [ ] Verify `setPrice` and round adjustment clamp to `[minimumPriceUpscaled, MAX_CURRENT_PRICE_UPSCALED]`
- [ ] Confirm `currentPrice()` matches `currentPriceUpScaled >> 10` at boundary values

### DIFF
--- a/src/PriceOracle.sol
+++ b/src/PriceOracle.sol
@@ -24,13 +24,8 @@ contract PriceOracle is AccessControl {
     // The number of the last round price adjusting happend
     uint64 public lastAdjustedRound;
 
-    // The minimum price allowed
-    uint32 public minimumPriceUpscaled = 24000 << 10; // we upscale it by 2^10
-
     // The priceBase to modulate the price
     uint32 public priceBase = 1048576;
-
-    uint64 public currentPriceUpScaled = minimumPriceUpscaled;
 
     // Constants used to modulate the price, see below usage
     uint32[9] public changeRate = [1049417, 1049206, 1048996, 1048786, 1048576, 1048366, 1048156, 1047946, 1047736];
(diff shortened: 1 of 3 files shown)


src/PriceOracle.sol
```diff
@@ -41,6 +41,11 @@ contract PriceOracle is AccessControl {
     // The length of a round in blocks.
     uint8 private constant ROUND_LENGTH = 152;
 
+    /// @dev Upper bound for upscaled price so `(currentPriceUpScaled >> 10)` fits in `uint32`.
+    /// Without this, `currentPrice()`'s `uint32(... >> 10)` truncates and can disagree with
+    /// `currentPriceUpScaled` and under-report vs `minimumPrice()`.
+    uint64 public constant MAX_CURRENT_PRICE_UPSCALED = uint64(uint256(type(uint32).max) << 10);
```
nugaon: I would have just cast once into uint64 but it is fine

src/PriceOracle.sol
```diff
@@ -182,6 +179,18 @@ contract PriceOracle is AccessControl {
     //            STATE READING           //
     ////////////////////////////////////////
 
+    /// @notice Clamp upscaled price to [minimumPriceUpscaled, MAX_CURRENT_PRICE_UPSCALED].
+    function _clampPriceUpscaled(uint64 priceUpScaled) private view returns (uint64) {
+        uint64 minU = uint64(minimumPriceUpscaled);
```
nugaon: this `minimumPriceUpscaled` should be a constant to save gas. `uint64 public constant MINIMUM_PRICE_UPSCALED = 24000 << 10;`

nugaon (approved):

PR #312, opened 2026-05-22
fix(redis): extract virtual randomness hook for fuzz harnesses

0xCardiE:
## Summary
- Make `updateRandomness` `internal virtual` and extract `_nextSeedValue` so test/fuzz wrappers can override post-reveal randomness without changing production semantics.

## Test plan
- [ ] Run existing Redistribution unit tests
- [ ] Confirm production `updateRandomness` behavior is unchanged (same keccak256 seed update)
- [ ] Verify fuzz harness subclasses can override `_nextSeedValue` when needed


0xCardiE:
Not needed.

PR #314, opened 2026-06-09
fix(signatures): reject ecrecover bypass

0xCardiE:
## Summary

Replaces raw `ecrecover` in `Signatures.sol` with OpenZeppelin `ECDSA.tryRecover` to close a signature verification bypass.

**Problem:** `recoverSigner` returned whatever `ecrecover` produced, including `address(0)` for malformed signatures, without validating `v` or rejecting high-`s` malleable signatures. In `socFunction`, the claimed signer comes from the submitted proof (`socProof[0].signer`), so an attacker could set `signer = address(0)` and supply a malformed signature; `socVerify` would then pass because `address(0) == address(0)`.

**Fix:** Introduce `verifySignature`, which uses `ECDSA.tryRecover` and only returns `true` when recovery succeeds with no error and the recovered address matches the expected signer. This also enforces valid `v`/`s` and rejects malleable signatures. Removes the hand-rolled `splitSignature` / `recoverSigner` path.

**Practical impact:** A full claim bypass would still require finding an `identifier` such that `keccak256(identifier, address(0))` equals the sampled `proveSegment`, which is not feasible in practice. The fix is still warranted for correctness and defense in depth — SOC authentication should not depend on keccak preimage difficulty.

## Changes

- `src/Util/Signatures.sol` — use OpenZeppelin `ECDSA.tryRecover` in `socVerify` and `postageVerify`
- `src/test/SignaturesHarness.sol` — thin wrapper for library unit tests
- `test/Signatures.test.ts` — regression tests for `address(0)` bypass, invalid length, malleable `s`, and valid signatures
- `test/Redistribution.test.ts` — claim-level test rejecting `address(0)` signer with malformed signature

## Test plan

- [x] `yarn compile`
- [x] `yarn test` (163 tests passing)
- [x] New unit tests cover the reported `address(0)` + malformed signature bypass
- [x] Existing SOC claim tests still pass (`wrong SOC signature`, successful SOC sampling claims)

### DIFF
--- a/src/Util/Signatures.sol
+++ b/src/Util/Signatures.sol
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.19;
 
-library Signatures {
-    error InvalidSignatureLength();
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 
+library Signatures {
     /** Hash of the message to sign */
     function getPostageMessageHash(
         bytes32 _chunkAddr,
@@ -25,7 +25,7 @@ library Signatures {
         bytes32 messageHash = getPostageMessageHash(_chunkAddr, _postageId, _index, _timeStamp);
         bytes32 ethMessageHash = getEthSignedMessageHash(messageHash);
 
-        return recoverSigner(ethMessageHash, _signature) == _signer;
+        return verifySignature(ethMessageHash, _signature, _signer);
     }
 
     function getEthSignedMessageHash(bytes32 _messageHash) internal pure returns (bytes32) {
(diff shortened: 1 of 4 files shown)


nugaon (approved):
computationally takes very long time to someone mine a SOC identifier so that its hash with `address(0)` results in the required proveSegment.

Regardless it does not change much, the code is fine and more simplified, we can use OpenZeppelin's ec recovery instead.

0xCardiE:
> computationally takes very long time to someone mine a SOC identifier so that its hash with `address(0)` results in the required proveSegment.
> 
> Regardless it does not change much, the code is fine and more simplified, we can use OpenZeppelin's ec recovery instead.

Thnx for review, added detailed description.

PR #316, opened 2026-06-13
fix(oracle): reject silent price update failures

0xCardiE:
## Summary

Fixes a bug where `PriceOracle.setPrice` / `adjustPrice` could silently fail while still advancing oracle state, desyncing from PostageStamp and blocking future price adjustments (including during claims).

- Update PostageStamp **before** committing oracle state (`currentPriceUpScaled`, `lastAdjustedRound`)
- Replace low-level `call` + `return false` with direct `IPostageStamp.setPrice` so failures revert explicitly
- Revert with `OraclePaused()` when paused instead of returning `false`
- Claims use a low-level `adjustPrice` call in `Redistribution` so winner selection still emits `PriceAdjustmentSkipped` when adjustment cannot run (e.g. oracle paused)

## Test plan

- [x] `yarn compile`
- [x] `yarn test` (159 passing)
- [x] New test: paused oracle reverts with `OraclePaused()`
- [x] New test: postage role revoked → revert and `lastAdjustedRound` unchanged

### DIFF
--- a/src/PriceOracle.sol
+++ b/src/PriceOracle.sol
@@ -47,13 +47,13 @@ contract PriceOracle is AccessControl {
      *@dev Emitted on every price update.
      */
     event PriceUpdate(uint256 price);
-    event StampPriceUpdateFailed(uint256 attemptedPrice);
 
     // ----------------------------- Custom Errors ------------------------------
     error CallerNotAdmin(); // Caller is not the admin
     error CallerNotPriceUpdater(); // Caller is not a price updater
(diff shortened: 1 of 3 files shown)


nugaon:
I think first the https://github.com/ethersphere/storage-incentives/pull/311/changes#diff-e93017bf979c1ba116cc1bfd4dd0c409ef62e4c8766ee26f5433822fb95ef4feR183 could be merged

